PR #10, opened 2018-07-26
Ins 40 instance domain

ilyapt:
**- What I did**
Instance Domain
Instance Domain Proxy
Instance Domain Factory
Tests

**- How to verify it**
Run tests


genesis/public/core/instance.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+// InstanceDomainName is a name for instance domain.
+const InstanceDomainName = "InstanceDomain"
+
+// InstanceDomain is a contract that stores instances of other domains
+type InstanceDomain interface {
+	// Base domain implementation.
+	domain.Domain
+	// CreateInstance is used to new instance as a child to domain storage.
```
megge-dream: You probably forgot word "create" here :) ("is used to **create** new instance")

genesis/public/core/instance_test.go
```diff
@@ -0,0 +1,267 @@
...
+	factory := NewInstanceDomainFactory()
+	proxy := factory.Create(parent)
+	instDomain, _ := newInstanceDomain(parent)
+
+	assert.Equal(t, &instanceDomainProxy{
+		instance: instDomain,
+	}, proxy)
+}
+
+func TestInstanceDomainFactory_CreateWithError(t *testing.T) {
```
LinnikD: There need to be two tests with errors:  
- with no parent
- with parent, whose AddChild returns error

LinnikD:
Ok

genesis/public/core/instance.go
```diff
@@ -0,0 +1,161 @@
...
+func (idf *instanceDomainFactory) GetClassID() string {
+	return class.InstanceDomainID
+}
+
+// GetReference returns nil for not published factory
+func (idf *instanceDomainFactory) GetReference() *object.Reference {
+	return nil
+}
+
+// Create factory method is used to create new InstanceDomain instances.
```
dmitry-panchenko: `Create is factory method that used...`
or
`Create - factory method...`

PR #11, opened 2018-07-27
INS-34: global resolver

megge-dream:
closes INS-34

**- What I did**
Resolver interface now gets and returns interface{}
Add InitiateResolver to smart contract & domain
Add tests for InitiateResolver
Fix register method for reference domain
Fixes of tests for reference domain

**- How to verify it**
Run tests

Add GlobalResolver instance

### DIFF
--- a/genesis/model/contract/smart_contract.go
+++ b/genesis/model/contract/smart_contract.go
@@ -22,12 +22,14 @@ import (
 	"github.com/insolar/insolar/genesis/mock/storage"
 	"github.com/insolar/insolar/genesis/model/class"
 	"github.com/insolar/insolar/genesis/model/object"
+	"github.com/insolar/insolar/genesis/model/resolver"
 )
 
 // SmartContract marks that object is smart contract.
 // TODO: Composite work interface
 type SmartContract interface {
-	object.Callable
+	object.Child
+	InitiateResolver(self interface{})
 }
 
 // BaseSmartContract is a base implementation of ComposingContainer, Callable and TypedObject interfaces.
@@ -37,16 +39,31 @@ type BaseSmartContract struct {
 	ChildStorage   storage.Storage
 	ContextStorage storage.Storage
 	Parent         object.Parent
+	Resolver       resolver.Resolver
 }
 
 // NewBaseSmartContract creates new BaseSmartContract instance with empty CompositeMap, ChildStorage and specific parent.
 func NewBaseSmartContract(parent object.Parent) *BaseSmartContract {
 	// TODO: NewCompositeHolder
-	return &BaseSmartContract{
+	sc := BaseSmartContract{
 		CompositeMap: make(map[string]object.Composite),
 		ChildStorage: storage.NewMapStorage(),
 		Parent:       parent,
 	}
+	sc.InitiateResolver(sc)
+	return &sc
+}
+
+// InitiateResolver create resolver.
+func (sc *BaseSmartContract) InitiateResolver(self interface{}) {
+	if sc.Resolver != nil {
+		return
+	}
+	scAsParent, ok := self.(object.Parent)
+	if !ok {
+		sc.Resolver = resolver.NewHandler(nil)
+	}
+	sc.Resolver = resolver.NewHandler(scAsParent)
 }
 
 // GetClassID return string representation of object's class.
--- a/genesis/model/contract/smart_contract.go
+++ b/genesis/model/contract/smart_contract.go
@@ -75,8 +75,8 @@ func (sc *BaseSmartContract) GetReference() *object.Reference {
 func (sc *BaseSmartContract) CreateComposite(compositeFactory object.CompositeFactory) (object.Composite, error) {
 	composite := compositeFactory.Create()
 	interfaceKey := composite.GetInterfaceKey()
-	_, isExist := sc.CompositeMap[interfaceKey]
-	if isExist {
+	_, exist := sc.CompositeMap[interfaceKey]
+	if exist {
 		return nil, fmt.Errorf("delegate with name %s already exist", interfaceKey)
 	}
 	sc.CompositeMap[interfaceKey] = composite
@@ -85,8 +85,8 @@ func (sc *BaseSmartContract) CreateComposite(compositeFactory object.CompositeFa
 
 // GetComposite return composite by its key (if its exist inside smart contract).
 func (sc *BaseSmartContract) GetComposite(key string) (object.Composite, error) {
-	composite, isExist := sc.CompositeMap[key]
-	if !isExist {
+	composite, exist := sc.CompositeMap[key]
+	if !exist {
 		return nil, fmt.Errorf("delegate with name %s does not exist", key)
 	}
 	return composite, nil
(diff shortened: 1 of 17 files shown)


genesis/model/contract/smart_contract.go
```diff
@@ -37,16 +39,31 @@ type BaseSmartContract struct {
...
 		CompositeMap: make(map[string]object.Composite),
 		ChildStorage: storage.NewMapStorage(),
 		Parent:       parent,
 	}
+	sc.InitiateResolver(sc)
+	return &sc
+}
+
+// InitiateResolver create resolver.
+func (sc *BaseSmartContract) InitiateResolver(self interface{}) {
```
LinnikD: InitResolver name will be better

genesis/model/contract/smart_contract.go
```diff
@@ -37,16 +39,31 @@ type BaseSmartContract struct {
...
+	if sc.Resolver != nil {
+		return
+	}
+	scAsParent, ok := self.(object.Parent)
+	if !ok {
```
LinnikD: NewHandler can handle this case. No need to pass self to method call as a parameter interface{}

genesis/model/contract/smart_contract.go
LinnikD: You can also rename it to GetResolver() and make resolver field private. GetResolver method will act like singleton - return instance or create it if necessary

genesis/model/resolver/child.go
```diff
@@ -47,7 +50,11 @@ func (r *childResolver) GetObject(ref *object.Reference, classID string) (object
 		return nil, fmt.Errorf("object is not Proxy")
 	}
 
-	if proxy.GetClassID() != classID {
+	classIDString, ok := classID.(string)
```
LinnikD: it is bad style to use types in names. Rename to classID. previous classID can be named like cls

genesis/model/resolver/context.go
```diff
@@ -51,19 +54,24 @@ func (r *contextResolver) GetObject(ref *object.Reference, classID string) (obje
 	}
 
 	for proxy.GetClassID() == class.ReferenceID {
-		contextHolderWithChildInterface, ok := contextHolder.(object.Child)
-		if !ok {
+		contextHolderWithChildInterface, isOk := contextHolder.(object.Child)
```
LinnikD: `ok` was better name

genesis/model/resolver/context.go
```diff
@@ -51,19 +54,24 @@ func (r *contextResolver) GetObject(ref *object.Reference, classID string) (obje
...
-		proxy, err = contextResolver.GetObject(proxy.(*object.Reference), classID)
+		newProxy, err := contextResolver.GetObject(proxy, classID)
 		if err != nil {
 			return nil, err
 		}
+		proxy = newProxy.(object.Proxy)
 	}
 
-	if proxy.GetClassID() != classID {
+	classIDString, ok := classID.(string)
```
LinnikD: simple names is better

genesis/model/resolver/global.go
```diff
@@ -24,31 +24,50 @@ import (
...
-	if proxy.GetClassID() != classID {
+	parent, ok := parentProxy.(object.Parent)
+	if !ok {
+		return nil, fmt.Errorf("object with domain `%s` can not have children", ref.Domain)
+	}
+	proxy, err := parent.GetChild(ref.Record)
+	if err != nil {
+		return nil, err
+	}
+	classIDString, ok := classID.(string)
```
LinnikD: here `classIDString ` also need to be renamed

genesis/model/resolver/handler.go
```diff
@@ -22,23 +22,28 @@ import (
...
+	return &Handler{
 		globalResolver:  GlobalResolver,
 		childResolver:   newChildResolver(p),
 		contextResolver: newContextResolver(p),
 	}
 }
 
 // GetObject resolve object by its reference and return its proxy.
-func (r *resolverHandler) GetObject(ref *object.Reference, classID string) (object.Proxy, error) {
+func (r *Handler) GetObject(reference interface{}, classID interface{}) (interface{}, error) {
```
LinnikD: Add description to doc: why we use interface{} types for reference, class and proxy

genesis/model/resolver/global.go
```diff
@@ -24,31 +24,50 @@ import (
...
+	if proxy.GetClassID() != classIDString {
 		return nil, fmt.Errorf("instance class is not `%s`", classID)
 	}
 	return proxy, nil
 }
 
+// SetMap set globalInstanceMap into globalResolver
+func (r *globalResolver) SetMap(globalInstanceMap *map[string]object.Proxy) error {
+	if r.globalInstanceMap != nil {
+		return fmt.Errorf("map was already set")
```
LinnikD: Is it really error case? If map already set we can act normally. Will it be error case in future implementation?

genesis/public/core/reference.go
```diff
@@ -33,13 +33,16 @@ type ReferenceDomain interface {
 	// Base domain implementation.
 	domain.Domain
 	// RegisterReference is used to publish new global references
-	RegisterReference(*object.Reference) (string, error)
+	RegisterReference(*object.Reference, string) (string, error)
 	// ResolveReference provides reference instance from record
 	ResolveReference(string) (*object.Reference, error)
+	// SetMap set globalResolverMap for references register/resolving
+	SetMap(globalInstanceMap *map[string]object.Proxy) error
```
LinnikD: `SetMap` is not good context name here. Maybe `InitGlobalResolver` is better?

genesis/model/resolver/context.go
megge-dream: declaration of ```ok``` shadows declaration at line 40 (in this function but outside of cycle)

LinnikD: `isChild` then ;)

genesis/public/core/reference.go
megge-dream: Its not about init global resolver, its about set inner reference domain map at the same memory point as global resolver map. InitGlobalMap mb?

genesis/public/core/reference.go
```diff
@@ -96,8 +117,8 @@ func newReferenceDomainProxy(parent object.Parent) *referenceDomainProxy {
 }
 
 // RegisterReference proxy call for instance method.
-func (rdp *referenceDomainProxy) RegisterReference(address *object.Reference) (string, error) {
-	return rdp.instance.RegisterReference(address)
+func (rdp *referenceDomainProxy) RegisterReference(address *object.Reference, classID string) (string, error) {
+	return rdp.instance.RegisterReference(address, classID)
 }
 
 // ResolveReference proxy call for instance method.
```
dmitry-panchenko: ResolveReference is a ...

genesis/public/core/reference.go
```diff
@@ -120,6 +141,11 @@ func (rdp *referenceDomainProxy) GetClassID() string {
 	return class.ReferenceDomainID
 }
 
+// InitGlobalMap proxy call for instance method.
```
dmitry-panchenko: InitGlobalMap is a

PR #17, opened 2018-07-31
INS-46: reference as interface

megge-dream:
closes INS-46

**- What I did**
Make Reference an interface

**- How to verify it**
Run tests

### DIFF
--- a/genesis/model/object/reference.go
+++ b/genesis/model/object/reference.go
@@ -42,18 +42,18 @@ type Reference interface {
 }
 
 type reference struct {
-	record string
 	domain string
+	record string
 	scope  ScopeType
 }
 
 // NewReference creates new reference instance.
-func NewReference(record string, domain string, scope ScopeType) (Reference, error) {
+func NewReference(domain string, record string, scope ScopeType) (Reference, error) {
 	switch scope {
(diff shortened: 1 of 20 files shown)


genesis/model/object/reference.go
```diff
@@ -33,45 +33,68 @@ const (
...
+
+type reference struct {
+	record string
+	domain string
+	scope  ScopeType
 }
 
 // NewReference creates new reference instance.
-func NewReference(record string, domain string, scope ScopeType) (*Reference, error) {
+func NewReference(record string, domain string, scope ScopeType) (Reference, error) {
```
LinnikD: You should pass domain as a 1st param. It is logically right

genesis/model/resolver/handler_test.go
```diff
@@ -17,13 +17,45 @@
...
+	"fmt"
 	"testing"
 
 	"github.com/insolar/insolar/genesis/mock/storage"
+	"github.com/insolar/insolar/genesis/model/class"
 	"github.com/insolar/insolar/genesis/model/object"
 	"github.com/stretchr/testify/assert"
 )
 
+type referenceNonExistentScope struct{}
```
LinnikD: invalidScopeReference may be better name

PR #27, opened 2018-08-03
INS-59: member domain

megge-dream:
closes INS-59

**- What I did**
Add MemberDomain interface, its implementation, proxy and factory.

**- How to verify it**
Run tests.

### DIFF
--- a/genesis/example/doc.go
+++ b/genesis/example/doc.go
@@ -1,0 +1,1 @@
+package example
--- a/genesis/example/doc.go
+++ b/genesis/example/doc.go
@@ -15,7 +15,7 @@
  */
 
 /*
-Package lottery provides smart contracts for building lottery for members, which can make bets from balance of their wallets.
+Package example provides smart contracts for building example of system work.
 
 Entities:
 
@@ -30,4 +30,4 @@ Entities:
 		memberProxy, err := mDomain.GetMember(record)
 
 */
-package lottery
+package example
--- a/genesis/example/doc.go
+++ b/genesis/example/doc.go
@@ -23,8 +23,8 @@ Entities:
 
 	Usage:
 
-		factory := NewMemberDomainFactory()
-		mDomain, error := factory.Create(parent)
+		factory := NewMemberDomainFactory(factoryParent)
+		mDomain, error := factory.Create(domainParent)
 
 		record, err := mDomain.CreateMember(someMemberFactory)
 		memberProxy, err := mDomain.GetMember(record)
--- a/genesis/example/doc.go
+++ b/genesis/example/doc.go
@@ -1,1 +1,33 @@
+/*
+ *    Copyright 2018 INS Ecosystem
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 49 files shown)


genesis/example/member_test.go
```diff
@@ -0,0 +1,335 @@
...
+}
+
+func TestNewMemberDomainProxy_WithNilParent(t *testing.T) {
+	_, err := newMemberDomainProxy(nil)
+	assert.EqualError(t, err, "parent must not be nil")
+}
+
+func TestMemberDomainProxy_CreateMember(t *testing.T) {
+	parent := &mockParent{}
+	proxy, err := newMemberDomainProxy(parent)
```
golangci: ineffectual assignment to `err`

genesis/example/member_test.go
```diff
@@ -0,0 +1,335 @@
...
+	member, err := proxy.CreateMember(factory)
+	assert.NoError(t, err)
+
+	_, err = uuid.FromString(member)
+	assert.NoError(t, err)
+}
+
+func TestMemberDomainProxy_GetMember(t *testing.T) {
+	parent := &mockParent{}
+	proxy, err := newMemberDomainProxy(parent)
```
golangci: ineffectual assignment to `err`

genesis/example/member.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/insolar/insolar/genesis/model/factory"
+	"github.com/insolar/insolar/genesis/model/object"
+	"github.com/insolar/insolar/genesis/model/resolver"
+)
+
+// MemberDomainName is a name for member domain.
+const MemberDomainName = "MemberDomain"
+
+// MemberDomainID is a string representation of class for MemberDomain interface implementations.
+const MemberDomainID = "MemberDomain"
```
LinnikD: why not in constant.go?

genesis/example/member.go
```diff
@@ -0,0 +1,163 @@
...
+	}, nil
+}
+
+// GetClassID returns string representation of MemberDomain's class.
+func (md *memberDomain) GetClassID() string {
+	return MemberDomainID
+}
+
+// CreateMember creates new member as a child to domain storage.
+func (md *memberDomain) CreateMember(fc factory.Factory) (string, error) {
```
LinnikD: CreateMember should do more specific things with factory. You need to create a factory of member class and set it into domain storage on domain creation. Then CreateMember method will only call MemberFactory inside

genesis/example/member.go
```diff
@@ -0,0 +1,163 @@
...
+	record, err := md.ChildStorage.Set(member)
+	if err != nil {
+		return "", err
+	}
+
+	return record, nil
+}
+
+// GetMember returns member from its record in domain storage.
+func (md *memberDomain) GetMember(record string) (resolver.Proxy, error) {
```
LinnikD: Here you also need to rely on MemberProxy class declaration,  not Proxy

genesis/example/member.go
megge-dream: Then Member class will be created, it will be done.

genesis/example/member.go
megge-dream: Then Member class will be created, it will be done.

genesis/example/doc.go
```diff
@@ -0,0 +1,33 @@
...
+	Usage:
+
+		factory := NewMemberDomainFactory(factoryParent)
+		mDomain, error := factory.Create(domainParent)
+
+		record, err := mDomain.CreateMember(someMemberFactory)
+		memberProxy, err := mDomain.GetMember(record)
+
+*/
+package example
```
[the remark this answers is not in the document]
LinnikD: let it stay =)

PR #37, opened 2018-08-03
INS-79: Add root record on DB init

andreyromancev:
closes INS-79

**- What I did**
Implemented initial storage record creation

**- How to verify it**
Run tests

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -3,92 +3,143 @@
 
 [[projects]]
   branch = "master"
+  digest = "1:26553bdecb5d2155e37185e91ac9e6cab762c06fdf231c151fb6d3ec60ff53b7"
+  name = "github.com/2tvenom/cbor"
+  packages = ["."]
+  pruneopts = "UT"
+  revision = "384e799bd841fc59220d0f5911b8cbde7b380875"
+
+[[projects]]
+  branch = "master"
+  digest = "1:d7e398c16b9c85f33ea98182809bfefe4e3ec8bf01f56de20f9721f9f7ab23de"
   name = "github.com/anacrolix/missinggo"
   packages = [
     ".",
     "inproc",
     "perf",
-    "pproffd"
+    "pproffd",
   ]
+  pruneopts = "UT"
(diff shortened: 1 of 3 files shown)


ledger/storage/leveldb/leveldb.go
```diff
@@ -96,9 +99,28 @@ func InitDB() (*LevelLedger, error) {
...
+	ledger := LevelLedger{
 		ldb: db,
-	}, nil
+		zeroRef: record.Reference{
+			Domain: record.ID{}, // TODO: fill domain
+			Record: zeroID,
+		},
+	}
+	_, err = db.Get([]byte(zeroRecordHash), nil)
+	switch err {
```
nordicdyno: @andreyromancev @ita-sammann Error checking in this way is not idiomatic go code. I would like to see this part rewritten

Gopkg.lock
```diff
@@ -102,31 +153,54 @@
...
+    "github.com/satori/go.uuid",
+    "github.com/spf13/pflag",
+    "github.com/stretchr/testify/assert",
+    "github.com/stretchr/testify/mock",
+    "github.com/syndtr/goleveldb/leveldb",
+    "github.com/syndtr/goleveldb/leveldb/comparer",
+    "github.com/syndtr/goleveldb/leveldb/opt",
+    "github.com/ugorji/go/codec",
+    "golang.org/x/crypto/sha3",
+  ]
```
nordicdyno: @andreyromancev @ita-sammann why so different format? Probably we should fix dep version for consistent packages versioning.

ita-sammann: @nordicdyno, it's not clear for me. These changes are no more in master and we can't reproduce this behaviour of `dep ensure`.

ita-sammann: Current dep version is 0.5.0 let's stick to it.

ledger/storage/leveldb/leveldb.go
ita-sammann: It makes sense. Let's use if/else if/else in similar cases.

PR #40, opened 2018-08-06
Ins 58 tests for class domain

ivanshibitov:
Add test for ClassDomain


genesis/public/core/class_test.go
```diff
@@ -29,5 +32,181 @@ func TestClassDomain_GetClassID(t *testing.T) {
...
 	assert.Equal(t, class.ClsDomainID, domainID)
 }
+
+func TestClassDomain_GetClass_NoSuchRecord(t *testing.T) {
+	parent := &mockParent{}
+	classDom, err := newClassDomain(parent)
+	assert.NoError(t, err)
+	cl, err := classDom.GetClass("test")
+
+	assert.Error(t, err)
```
megge-dream: It's better to use `assert.EqualError(t, err, <some_error_message>)` and check error message too.

genesis/public/core/class_test.go
```diff
@@ -29,5 +32,181 @@ func TestClassDomain_GetClassID(t *testing.T) {
...
+	classDom, err := newClassDomain(parent)
+
+	assert.NoError(t, err)
+
+	recordId, regErr := classDom.RegisterClass(NewClassDomainFactory(parent))
+
+	assert.NoError(t, regErr)
+	assert.NotEmpty(t, recordId)
+
+	_, err = uuid.FromString(recordId)
```
ilyapt: uuid.FromString returns error for empty recordId, NotEmpty check is redundantly

PR #47, opened 2018-08-07
Ins 105 make wallet domain proxy and factory

ivanshibitov:
**- What I did**
Add Base code for WalletDomain Proxy & Factory

### DIFF
--- a/genesis/example/wallet.go
+++ b/genesis/example/wallet.go
@@ -114,7 +114,7 @@ type walletDomainFactory struct {
 }
 
 func (wdf *walletDomainFactory) Create(parent object.Parent) (resolver.Proxy, error) {
-	proxy, err := newMemberDomainProxy(parent)
+	proxy, err := newWalletDomainProxy(parent)
 	if err != nil {
 		return nil, err
 	}
--- a/genesis/example/wallet.go
+++ b/genesis/example/wallet.go
@@ -26,7 +26,7 @@ import (
 	"github.com/insolar/insolar/genesis/model/resolver"
 )
 
-const WalletDomainName = "WaletDomain"
+const WalletDomainName = "WalletDomain"
 
 type WalletDomain interface {
 	domain.Domain
@@ -128,6 +128,7 @@ func (wdf *walletDomainFactory) Create(parent object.Parent) (resolver.Proxy, er
 }
 
 func (wdf *walletDomainFactory) GetParent() object.Parent {
+	// TODO: return real parent, fix tests
 	return nil
 }
 
--- a/genesis/example/wallet.go
+++ b/genesis/example/wallet.go
@@ -1,0 +1,142 @@
+/*
+ *    Copyright 2018 INS Ecosystem
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


genesis/example/wallet.go
```diff
@@ -0,0 +1,151 @@
...
+	record, err := wd.ChildStorage.Set(wallet)
+	if err != nil {
+		return "", err
+	}
+
+	return record, nil
+}
+
+func (wd *walletDomain) GetWallet(record string) (resolver.Proxy, error) {
+	wallet, err := wd.ChildStorage.Get(record)
```
megge-dream: Its better to use `wd.GetChild(record)`

genesis/example/wallet_test.go
```diff
@@ -0,0 +1,205 @@
...
+	_, err := newWalletDomain(nil)
+	assert.EqualError(t, err, "parent must not be nil")
+}
+
+func TestNewWalletDomain(t *testing.T) {
+	parent := &mockParent{}
+	proxy, err := newWalletDomain(parent)
+
+	assert.NoError(t, err)
+	assert.Equal(t, proxy,
```
megge-dream: In `assert.Equal` second param is expected value, and third - actual value. And fix indents (last should be at next line).

genesis/example/wallet_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+func TestWalletDomain_GetClassID(t *testing.T) {
+	parent := &mockParent{}
+	wdomain, err := newWalletDomain(parent)
+
+	assert.NoError(t, err)
+	assert.Equal(t, class.WalletDomainID, wdomain.GetClassID())
+}
+
+func TestWalletDomain_GetWallet_NoShuchRecord(t *testing.T) {
```
megge-dream: May be _No**Such**Record_?

genesis/example/wallet.go
```diff
@@ -0,0 +1,151 @@
...
+	return &walletDomain{
+		BaseDomain: *domain.NewBaseDomain(parent, WalletDomainName),
+	}, nil
+}
+
+func (wd *walletDomain) GetClassID() string {
+	return class.WalletDomainID
+}
+
+func (wd *walletDomain) CreateWallet(fc factory.Factory) (string, error) {
```
megge-dream: Add comment to this method

genesis/example/wallet.go
```diff
@@ -0,0 +1,151 @@
...
+
+	record, err := wd.ChildStorage.Set(wallet)
+	if err != nil {
+		return "", err
+	}
+
+	return record, nil
+}
+
+func (wd *walletDomain) GetWallet(record string) (resolver.Proxy, error) {
```
megge-dream: Add comment to this method

genesis/example/wallet.go
LinnikD: In futher implementation you should pass member to this method

genesis/example/wallet.go
LinnikD: Do we really need GetWallet method? Can i get another person wallet and act with it?

ivanshibitov: I add comment to method in interface

genesis/example/wallet.go
ivanshibitov: I've already done it for interface method

PR #49, opened 2018-08-08
INS-60: member contract

megge-dream:
closes INS-60

**- What I did**
Add Member smart contract and fix MemberDomain to create and get Member interface instances

**- How to verify it**
Run tests


genesis/example/member.go
```diff
@@ -17,134 +17,140 @@
...
-	// Base domain implementation.
-	domain.Domain
-	// CreateMember is used to create new member as a child to domain storage.
-	CreateMember(factory.Factory) (string, error)
-	// GetMember returns member from its record in domain storage.
-	GetMember(string) (resolver.Proxy, error)
+type member struct {
+	contract.BaseSmartContract
+	Username   string
+	privateKey *rsa.PrivateKey
```
LinnikD: no private key stored in blockchain, it is only passed when node is going to be initialized

genesis/example/member.go
```diff
@@ -17,134 +17,140 @@
...
-// CreateMember is a proxy call for instance method.
-func (mdp *memberDomainProxy) CreateMember(fc factory.Factory) (string, error) {
-	return mdp.Instance.(MemberDomain).CreateMember(fc)
+// GetUsername is a proxy call for instance method.
+func (mp *memberProxy) GetUsername() string {
+	return mp.Instance.(Member).GetUsername()
+}
+
+// GetPublicKey is a proxy call for instance method.
+func (mp *memberProxy) GetPublicKey() *rsa.PublicKey {
```
LinnikD: CreateComposite and GetComposite is not a public methods

genesis/example/member_domain.go
```diff
@@ -0,0 +1,187 @@
...
+
+// GetClassID returns string representation of MemberDomain's class.
+func (md *memberDomain) GetClassID() string {
+	return class.MemberDomainID
+}
+
+// CreateMember creates new member as a child to domain storage.
+func (md *memberDomain) CreateMember() (string, error) {
+	// Get child by memberFactoryRecord
+	child, err := md.GetChild(md.memberFactoryRecord)
```
LinnikD: This should be done using childResolver, GetChild method should not be public anymore

genesis/example/member_domain.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	// Create member
+	member, err := mf.Create(md)
+	if err != nil {
+		return "", err
+	}
+	if member == nil {
+		return "", fmt.Errorf("factory returns nil")
+	}
+	_, ok = member.(Member)
```
LinnikD: type checks should be resolver responsibility. Any ideas how we can do this?

genesis/example/member_domain.go
```diff
@@ -0,0 +1,187 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	return record, nil
+}
+
+// GetMember returns member from its record in domain storage.
+func (md *memberDomain) GetMember(record string) (Member, error) {
+	member, err := md.GetChild(record)
```
LinnikD: also use childResolver instead

genesis/example/member.go
megge-dream: No, but for now we should implement all interface methods in proxy object

ilyapt: I think I'll do this in my job

genesis/example/member_domain.go
megge-dream: We still should convert types manually, but resolver will give us error about wrong type before this line, so here check is redundant

PR #52, opened 2018-08-09
INS-97: public interface arrived

ilyapt:
**- What I did**

**- How I did it**

**- How to verify it**


genesis/model/object/callable.go
```diff
@@ -19,22 +19,4 @@ package object
 // Callable allows itself to be called by its reference.
```
megge-dream: This comment should be changed

ilyapt: Any ideas?

LinnikD: Callable is a marker interface that shows that object is able to be called

genesis/example/member_domain_test.go
```diff
@@ -99,55 +98,42 @@ func (p *mockParentWithError) AddChild(child object.Child) (string, error) {
...
-	return "mockProxy"
-}
-
-func (p *mockProxy) GetParent() object.Parent {
-	return p.parent
-}
-
-type mockFactory struct {
-	mockCallable
+type mockFactoryProxy struct {
```
megge-dream: Where is no need to name this variable `mockFactoryProxy`, because Factory is always a Proxy. `mockFactory` is more clear in this case.

ilyapt: Sure

genesis/model/object/reference.go
```diff
@@ -45,7 +45,7 @@ type Reference interface {
 // TODO: ProxyReference & ReferenceInterface
 
 type reference struct {
```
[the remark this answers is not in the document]
ilyapt: Will be done in next sprint

genesis/model/resolver/proxy.go
```diff
@@ -17,40 +17,38 @@
...
-
-// SetReference is a proxy call for instance method.
-func (bp *BaseProxy) SetReference(reference object.Reference) {
-	bp.Instance.SetReference(reference)
+// GetClassID is a proxy call for instance method.
+func (bp *BaseProxy) GetClassID() string {
+	return class.ProxyID
 }
 
 // GetParent is a proxy call for instance method.
```
megge-dream: Now `GetParent` is not a proxy call, its `always returns nil`.

PR #54, opened 2018-08-09
INS-122: composite mechanism

megge-dream:
closes INS-122

**- What I did**
Add new composite to ChildStorage of ComposingContainer

**- How to verify it**
Run tests


genesis/model/contract/smart_contract.go
```diff
@@ -76,7 +76,18 @@ func (sc *BaseSmartContract) CreateComposite(compositeFactory object.CompositeFa
 	if exist {
 		return nil, fmt.Errorf("delegate with name %s already exist", interfaceKey)
 	}
+
+	compositeAsChild, ok := composite.(object.Child)
```
LinnikD: Why composite is not a child by default?

megge-dream: Because we separate responsibility of interfaces. Composite about ability to be injected into smth, Child - about having a parent. If Composite also must have parent, than Composite interface must require GetParent() method specifically.

PR #61, opened 2018-08-10
Ins-134: add domain to record keys

nordicdyno:
**- What I did**

* switch in storage on key generation from record.Reference (to avoid conflicts for records in different domains)

* storage key generation from record.Reference to avoid conflicts for records in different domains


ledger/artifactmanager/artifactmanager.go
```diff
@@ -35,8 +35,8 @@ func (m *LedgerArtifactManager) checkRequestRecord(requestID record.ID) error {
 	return nil
 }
 
-func (m *LedgerArtifactManager) getCodeRecord(codeID record.ID) (*record.CodeRecord, error) {
-	rec, err := m.storer.GetRecord(codeID)
+func (m *LedgerArtifactManager) getCodeRecord(codeRef record.Reference) (*record.CodeRecord, error) {
```
andreyromancev: Maybe we should pass codeRef by reference here as well?
Same for other methods like getCodeRecordCode

ledger/artifactmanager/artifactmanager.go
```diff
@@ -61,47 +61,47 @@ func (m *LedgerArtifactManager) getCodeRecordCode(codeID record.ID) ([]byte, err
 }
 
 func (m *LedgerArtifactManager) storeRecord(rec record.Record) (record.Reference, error) {
```
andreyromancev: Return reference here instead of value?

ledger/artifactmanager/artifactmanager.go
```diff
@@ -355,16 +355,16 @@ func (m *LedgerArtifactManager) DeactivateObj(requestRef, objRef record.Referenc
 			HeadRecord: objRef,
 			AmendedRecord: record.Reference{
```
andreyromancev: Should be just `AmendedRecord: objIndex.LatestStateRef`. No need to create another Record.

ledger/artifactmanager/artifactmanager.go
```diff
@@ -396,7 +396,7 @@ func (m *LedgerArtifactManager) UpdateObj(
...
```
andreyromancev: Here as well, just `AmendedRecord: objIndex.LatestStateRef`

PR #65, opened 2018-08-14
INS-98: Common package

ilyapt:
**- What I did**
Joined object, reference and factory into object package.

**- How to verify it**
Run tests.


genesis/model/object/context.go
```diff
@@ -14,30 +14,29 @@
  *    limitations under the License.
  */
 
-package resolver
+package object
```
LinnikD: For now you need to show that it is resolver context.   Rename file to resolver.go and set there Resolver interface and all 3 implementation.

PR #69, opened 2018-08-15
INS-135: Make common error objects.

andreyromancev:
Closes INS-135

**- What I did**
Moved common errors to vars.


ledger/artifactmanager/errors.go
```diff
@@ -0,0 +1,11 @@
+package artifactmanager
+
+import "github.com/pkg/errors"
```
nordicdyno: should be a list according to code style

PR #72, opened 2018-08-15
Ins 172 allowance smart contract

ivanshibitov:
**- What I did**
1) add Allowance 
2) Add tests

**- How to verify it**
Run tests


genesis/example/wallet.go
```diff
@@ -47,6 +47,7 @@ func newWallet(parent object.Parent) (Wallet, error) {
 }
 
 func (w *wallet) GetBalance() int {
+	//w.GetOrCreateComposite()
```
ilyapt: Commented code should be removed

genesis/example/allowance.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/insolar/insolar/genesis/model/class"
+	"github.com/insolar/insolar/genesis/model/contract"
+	"github.com/insolar/insolar/genesis/model/object"
+)
+
+type Allowance interface {
+	object.Composite
+	contract.SmartContract
+	GetAmount() int
+	GetSender() string
```
LinnikD: You need to store both sender and recipient

LinnikD: You need an operation that marks Allowance as completed.

LinnikD: Why sender is string?

ivanshibitov: About recipient: I add it, but it's not pretty clear now if we need it. Since, allowance may locate inside recipient, and that is why we might not need this field

genesis/example/allowance.go
```diff
@@ -29,13 +29,16 @@ type Allowance interface {
 	contract.SmartContract
 	GetAmount() int
 	GetSender() string
+	GetReciever() string
+	MarkCompleted()
```
megge-dream: You probably need `IsCompleted()` method as well, so we can check only not completed allowances later for getting balance.

PR #75, opened 2018-08-16
INS-131 encrypted transport

AndreyBronin:
**- What I did**
Transport messages encryption possibility was added.
two transport bug fixed

**- How I did it**
By new transport implementation based on KCP protocol

**- How to verify it**
unit tests


network/host/transport/factory.go
```diff
@@ -26,3 +26,29 @@ import (
...
+	return &utpTransportFactory{}
+}
+
+// Create creates new UTP Transport.
+func (utpTransportFactory *utpTransportFactory) Create(conn net.PacketConn, proxy relay.Proxy) (Transport, error) {
+	return NewUTPTransport(conn, proxy)
+}
+
+type kcpTransportFactory struct{}
+type kcpSecureOptions struct {
```
golangci: `kcpSecureOptions` is unused

network/host/message/message.go
```diff
@@ -140,15 +141,13 @@ func DeserializeMessage(conn io.Reader) (*Message, error) {
 		return nil, err
 	}
 
-	msgBytes := make([]byte, length)
-	_, err = conn.Read(msgBytes)
-	if err != nil {
-		return nil, err
+	var reader bytes.Buffer
+	for uint64(reader.Len()) < length {
+		reader.ReadFrom(conn)
```
golangci: Error return value of `reader.ReadFrom` is not checked

network/host/transport/kcptransport.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+
+	_, err = session.Write(data)
+
+	return err
+}
+
+func (t *kcpTransport) handleAcceptedConnection(session *kcp.UDPSession) {
+	for {
+		session.SetDeadline(time.Now().Add(time.Millisecond * 50))
```
golangci: Error return value of `session.SetDeadline` is not checked

PR #82, opened 2018-08-22
Ins 190 messagerouter network integration

AndreyBronin:
**- What I did**
messagerouter.Route method which uses DHT network to deliver and run messages on reference node

**- How I did it**
using Serialization\Deserialization to network RPC

**- How to verify it**
 tests with real network DHT node

### DIFF
--- a/messagerouter/messagerouter.go
+++ b/messagerouter/messagerouter.go
@@ -19,7 +19,6 @@ package messagerouter
 import (
 	"bytes"
 	"encoding/gob"
-	"log"
 
 	"github.com/insolar/insolar/network/host"
 	"github.com/insolar/insolar/network/host/id"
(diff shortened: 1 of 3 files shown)


messagerouter/messagerouter.go
```diff
@@ -39,16 +52,75 @@ type Message struct {
...
+	if err != nil {
+		return nil, err
+	}
+
 	data, res, err := r.LogicRunner.Execute(msg.Reference, msg.Method, msg.Arguments)
-	return Response{data, res}, err
+	return Serialize(Response{data, res, err})
+}
+
+func (r *MessageRouter) getNodeID(reference string) (nodeId id.ID) {
```
golangci: method result nodeId should be nodeID

PR #93, opened 2018-08-28
INS-196 Node level network

ooki2day:
**- What I did**

Added a new layer between DHT and MessageRouter - nodenetwork.Node

**- How to verify it**

See node_test.go


network/nodenetwork/node.go
```diff
@@ -0,0 +1,118 @@
...
+	_, err := node.dht.RemoteProcedureCall(node.createContext(), targetID, method, args)
+	return err
+}
+
+// RecvRPC returns RPC request.
+func (node Node) RecvRPC() <-chan *packet.RequestDataRPC {
+	return node.receivedRPC
+}
+
+func (node Node) listenDHT(start, stop chan bool) {
```
golangci: func `Node.listenDHT` is unused

network/hostnetwork/packet/request.go
```diff
@@ -32,6 +32,11 @@ const (
 	RevokeAuth
 )
 
+// RequestCheckNodePriv is data for check node privileges.
+type RequestCheckNodePriv struct {
+	PrivKey string
```
AndreyBronin: comment what is  PrivKey, somebody may think it's private key instead privilegie key

PR #99, opened 2018-08-28
Ins 229 forge imports

Vany:
**Use real/fake foundation dualism**
Now preprocessor can generate contracts with forged (server) foundation library
this library created and moved to goplugin


logicrunner/goplugin/preprocessor/main.go
```diff
@@ -435,3 +457,38 @@ func generateMethodProxy(parsed *parsedFile, method *ast.FuncDecl) string {
 
 	return text
 }
+
+func cmdRewriteImports(fname string, w io.Writer) error {
+	parsed, err := parseFile(fname)
+	if err != nil {
+		return errors.Wrap(err, "couldn't parse")
+	}
+	rewriteImports(parsed)
```
golangci: Error return value of `rewriteImports` is not checked

PR #101, opened 2018-08-29
hottest fix node tests

ooki2day:
**- What I did**

hot fix for node tests

### DIFF
--- a/network/nodenetwork/node.go
+++ b/network/nodenetwork/node.go
@@ -33,19 +33,17 @@ type Node struct {
 }
 
 // NewNode creates a node with given args.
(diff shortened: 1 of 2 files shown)


network/nodenetwork/node.go
```diff
@@ -33,19 +33,17 @@ type Node struct {
...
-		role:     role,
 	}
-	err := node.dht.CheckNodeRole(node.createContext(), node.domainID)
-	if err != nil {
-		return nil, err
-	}
-	return node, err
+}
+
+func (node *Node) setRole(newRole string) {
```
golangci: func `(*Node).setRole` is unused

PR #131, opened 2018-09-04
Ins 294 ledger in logic runner

ita-sammann:
**- What I did**
Acquired ArtifactManager from Ledger in LogicRunner. Not from core.Components.


Vany:
Yes, it is it. Thanx.

PR #134, opened 2018-09-04
Ins 255 support any sc package name

ivanshibitov:
**- What I did**
Add option 'package' to preprocessor to change package name to 'main'

**- How to verify it**
Run tests


ivanshibitov:
No need in this task

PR #147, opened 2018-09-06
INS-209: plugin compiler moved into preprocessor

ilyapt:
**- What I did**
- Moved main func from preprocessor to /cmd/icc
- Added contracts compile code from goplugin_test to icc
- Fixed tests etc


logicrunner/goplugin/goplugin_test.go
```diff
@@ -97,7 +101,7 @@ func generateContractProxy(root string, name string) error {
 
 	contractPath := root + "/src/contract/" + name + "/main.go"
 
-	out, err := exec.Command("./preprocessor/preprocessor", "proxy", "-o", dstDir+"/main.go", "--code-reference", "testReference", contractPath).CombinedOutput()
+	out, err := exec.Command("../../cmd/icc/icc", "proxy", "-o", dstDir+"/main.go", "--code-reference", "testReference", contractPath).CombinedOutput()
```
ivanshibitov: "../../cmd/icc/icc" - is repeating several times. It should be brought out to some variable

PR #190, opened 2018-09-12
INS-384: Add method for getting jet ref from record ref.

andreyromancev:
Closes INS-384

### DIFF
--- a/ledger/jetcoordinator/jetcoordinator.go
+++ b/ledger/jetcoordinator/jetcoordinator.go
@@ -19,6 +19,7 @@ package jetcoordinator
 import (
 	"fmt"
 
+	"github.com/insolar/insolar/core"
 	"github.com/insolar/insolar/ledger/jetdrop"
 	"github.com/insolar/insolar/ledger/record"
 	"github.com/insolar/insolar/ledger/storage"
@@ -26,7 +27,8 @@ import (
 
 // JetCoordinator is responsible for all jet interactions
 type JetCoordinator struct {
-	db *storage.DB
+	db          *storage.DB
+	rootJetNode *JetNode
 }
 
 // Pulse creates new jet drop and ends current slot.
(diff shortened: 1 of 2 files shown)


ledger/jetcoordinator/jetnode.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+type JetNode struct {
+	threshold uint64
+	ref       *core.RecordRef
+	left      *JetNode
+	right     *JetNode
+}
+
+func (jn *JetNode) GetContaining(objRef uint64) *core.RecordRef {
```
ita-sammann: Shall it be 64 bytes, not bits?

ledger/jetcoordinator/jetnode.go
```diff
@@ -0,0 +1,28 @@
...
+	}
+
+	switch bytes.Compare(objRef[:], jn.ref[:]) {
+	case 0:
+		return &jn.ref
+	case -1:
+		return jn.left.GetContaining(objRef)
+	default:
+		return jn.right.GetContaining(objRef)
+	}
```
ita-sammann: Why 3 cases? Only leafs of tree should represent Jets

andreyromancev: case 0: its jet itself
The rest two are for branches.

ledger/jetcoordinator/jetcoordinator.go
```diff
@@ -107,7 +109,23 @@ func (jc *JetCoordinator) getNextValidators(candidates [][]byte, count int) ([][
...
 
+func (jc *JetCoordinator) jetRef(objRef core.RecordRef) *core.RecordRef { // nolint: megacheck
+	return jc.rootJetNode.GetContaining(&objRef)
+}
+
 // NewJetCoordinator creates new coordinator instance.
 func NewJetCoordinator(db *storage.DB) (*JetCoordinator, error) {
-	return &JetCoordinator{db: db}, nil
+	// TODO: temporary jet tree (later will be fetched fom config or ledger)
+	rootJetNode := &JetNode{
```
ita-sammann: Same thing as below. Not all tree nodes are Jets

andreyromancev: They used here as thresholds. If node doesn't have children it leaf, then we use it as ref. In other cases we treat it like splitting value.

PR #191, opened 2018-09-12
Ins 373 pulsar daemon

egorikas:
**- What I did**
I created a draft prototype of the pulsar. In the nearest future I am going to extend it and remove some mock code like `handleRequest`
**- How I did it**

**- How to verify it**

Pulsar's prototype

### DIFF
--- a/pulsar/entropygenerator.go
+++ b/pulsar/entropygenerator.go
@@ -18,14 +18,16 @@ package pulsar
 
 import "crypto/rand"
 
+type Entropy [8]byte
+
 type EntropyGenerator interface {
-	GenerateEntropy() [8]byte
+	GenerateEntropy() Entropy
 }
 
 type StandardEntropyGenerator struct {
 }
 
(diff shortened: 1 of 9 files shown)


pulsar/pulse.go
```diff
@@ -0,0 +1,29 @@
...
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+package pulsar
+
+//Base data struct for a pulse
+type Pulse struct {
+	PulseNumber uint64
+	Entropy     [8]byte
```
dmitry-panchenko: I think there should be an alias for `[8]byte`

ita-sammann: And it must be 64 byte.
Let's make `type PulseEntropy [64]byte`

pulsar/entropygenerator.go
```diff
@@ -18,14 +18,16 @@ package pulsar
 
 import "crypto/rand"
 
+type Entropy [8]byte
```
ita-sammann: It must be 64 bytes, not 64 bits

PR #197, opened 2018-09-13
INS-309: mock jet coordinator

dmitry-panchenko:
**- What I did**
Added mock jet coordinator

**- How I did it**
With config file

**- How to verify it**
Unknown but run tests

Added jet coordinator


core/ledger.go
```diff
@@ -16,8 +16,181 @@
...
+	// IsAuthorized checks for role on concrete pulse for the address
+	IsAuthorized(role JetRole, obj RecordRef, pulse PulseNumber, node RecordRef) bool
+
+	// TODO: add docs
+	QueryRole(role JetRole, obj RecordRef, pulse PulseNumber) []RecordRef
+}
+
+type PulseManager interface {
+	// TODO: docs
+	Current() (PulseNumber, error)
```
andreyromancev: Should we use `Pulse` here as well?

dmitry-panchenko: Yep.

PR #218, opened 2018-09-16
INS-341 network traces

AndreyBronin:
**- What I did**
Add more traces to network subsystem.
Add formatted output methods to log package.
Cmd apps log switched to Insolar log package.

**- How to verify it**
unit tests


core/logger.go
```diff
@@ -25,29 +25,41 @@ type Logger interface {
 	Debug(...interface{})
 	// Debugln logs a message at level Debug.
 	Debugln(...interface{})
+	// Debugf formatted logs a message at level Debug.
+	Debugf(string, ...interface{})
 
 	// Info logs a message at level Info.
 	Info(...interface{})
 	// Infoln logs a message at level Info.
 	Infoln(...interface{})
```
ivanshibitov: Do we really need *ln() versions of logging?
May be we should remove them and make other versions to print new line implicitly and obligatory?

network/hostnetwork/dht.go
```diff
@@ -402,7 +400,7 @@ func (dht *DHT) iterateBootstrapHosts(
 			if err != nil {
 				continue
 			}
-			log.Println("sending ping to " + bn.Address.String())
+			log.Debugln("sending ping to " + bn.Address.String())
```
ivanshibitov: Almost everywhere use write to log only warnings and error, and as a result you don't have kind of "access log" ( you have it only in debug mode. ). But normally, system will work without debug mode, and it will be hard to understand without "access log".
One more thing - I think you log something like "query id". It's very helpful thing

core/logger.go
AndreyBronin: Debug() and Debugln() has the same behavior, but interface has both methods for compatibility

network/hostnetwork/dht.go
AndreyBronin: Default log level is Info, debug messages added for debugging and may be will removed in the future

core/logger.go
ivanshibitov: If behaviors are the same, it's a little bit confusing to have two functions

network/hostnetwork/dht.go
ivanshibitov: Yes, I mean, that you don't have information about system until everything is wrong

AndreyBronin: we also have metrics counters which shows that system is working

PR #222, opened 2018-09-17
Benchmark for contract calling

Vany:
**Benchmark for contracts**

### DIFF
--- a/ledger/ledgertestutil/ledgertestutil.go
+++ b/ledger/ledgertestutil/ledgertestutil.go
@@ -28,7 +28,7 @@ import (
 
 // TmpLedger crteates ledger on top of temporary database.
 // Returns *ledger.Ledger andh cleanup function.
-func TmpLedger(t *testing.T, dir string) (*ledger.Ledger, func()) {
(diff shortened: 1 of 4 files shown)


ledger/ledgertestutil/ledgertestutil.go
```diff
@@ -28,7 +28,7 @@ import (
 
 // TmpLedger crteates ledger on top of temporary database.
 // Returns *ledger.Ledger andh cleanup function.
-func TmpLedger(t *testing.T, dir string) (*ledger.Ledger, func()) {
+func TmpLedger(t testing.TB, dir string) (*ledger.Ledger, func()) {
```
ruz: Don't see reason for this change? What is the point?

Vany: we want use helpers in benchmarks

ruz: Sorry, I was not clear, but what is the point to switch from pointer to not pointer? T -> TB change is perfectly clear.

PR #245, opened 2018-09-19
INS-452: Message to Event

dmitry-panchenko:
**- What I did**
Renamed messages to events


eventbus/event/base.go
```diff
@@ -0,0 +1,114 @@
...
+func (baseEvent) Serialize() (io.Reader, error) {
+	panic("Do not use base")
+}
+
+func (baseEvent) GetReference() core.RecordRef {
+	panic("Do not use base")
+}
+
+// EventType is a enum type of event
+type EventType byte
```
golangcibot: type name will be used as event.EventType by other packages, and that stutters; consider calling this Type

PR #269, opened 2018-09-20
Version 0.3.0

ita-sammann:
**- What I did**
~Hacked~ Prepared everything for networking demo


bootstrap/bootstrap.go
```diff
@@ -75,23 +75,27 @@ func (b *Bootstrapper) Start(c core.Components) error {
...
-	isLightExecutor, err := jc.IsAuthorized(core.RoleLightExecutor, *am.RootRef(), currentPulse.PulseNumber, nodeID)
-	if err != nil {
-		return errors.Wrap(err, "couldn't authorized node")
-	}
-	if !isLightExecutor {
-		return nil
-	}
+	// TODO: Denis Linnik
+	// Temporary disable role checking.
+	// Because of bad API module which gets contracts' addresses from bootstrap.
```
LinnikD: Solution is to make bootstrap work with different roles. API work properly now

bootstrap/bootstrap.go
```diff
@@ -64,34 +65,43 @@ func (b *Bootstrapper) Start(c core.Components) error {
...
 	if rootRefChildren.HasNext() {
 		rootDomainRef, err := rootRefChildren.Next()
 		if err != nil {
+			log.Errorln("Couldn't get next child of RootRef object")
 			return errors.Wrap(err, "couldn't get next child of RootRef object")
 		}
 		b.rootDomainRef = &rootDomainRef
+		log.Debugf("Root domain ref: %s", b.rootDomainRef)
 		return nil
+	} else {
```
golangcibot: if block ends with a return statement, so drop this else and outdent its block

PR #275, opened 2018-09-21
INS-397: tests for proxy generation

ilyapt:
**- What I did**
- wrote tests for proxy generation and showing their diff

**- How to verify it**
- run tests

### DIFF
--- a/cmd/insgocc/insgocc.go
+++ b/cmd/insgocc/insgocc.go
@@ -24,7 +24,6 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
-	"strings"
 
 	"github.com/insolar/insolar/logicrunner/goplugin/preprocessor"
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


logicrunner/logicrunner_test.go
```diff
@@ -905,3 +909,35 @@ func (c *Child) GetNum() int {
 		assert.Equal(b, []interface{}([]interface{}{uint64(5)}), r)
 	}
 }
+
+func TestProxyGeneration(t *testing.T) {
```
dmitry-panchenko: FYI: for better IDE support you could use table tests here instead of loops

logicrunner/logicrunner_test.go
```diff
@@ -905,3 +909,37 @@ func (c *Child) GetNum() int {
...
+	contracts := []string{
```
LinnikD: We need to check all contracts in dir experiment

cmd/insgocc/insgocc.go
```diff
@@ -64,21 +63,6 @@ func (r *outputFlag) Type() string {
 	return "file"
 }
 
-func findPath() (string, error) {
-	gopath := os.Getenv("GOPATH")
-	if gopath == "" {
```
nordicdyno: I think, this is the pretty legit case and should not be an error.
Probably it's a good idea to use build.Default from `go/build` package (https://golang.org/pkg/go/build/#Context)

PR #325, opened 2018-10-01
Noissue crash 01

ivanshibitov:
**- What I did**

**- How I did it**


ivanshibitov:
Fixed vy update to go1.11

PR #343, opened 2018-10-04
INS-540: add seedmanager

ivanshibitov:
**- What I did**
1) Add seed pool
2) Add get_seed request to api

**- How to verify it**
Run tests

### DIFF
--- a/api/request_handler.go
+++ b/api/request_handler.go
@@ -307,6 +307,7 @@ func (rh *RequestHandler) ProcessIsAuthorized() (map[string]interface{}, error) 
 	return result, nil
 }
 
+// ProcessGetSeed processes get seed request
 func (rh *RequestHandler) ProcessGetSeed() (map[string]interface{}, error) {
 	result := make(map[string]interface{})
 	seed, err := rh.seedGenerator.Next()
--- a/api/request_handler.go
+++ b/api/request_handler.go
@@ -17,10 +17,12 @@
 package api
 
 import (
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"reflect"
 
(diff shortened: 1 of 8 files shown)


api/seedmanager/seedgenerator.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func init() {
+	rand.Seed(time.Now().UTC().UnixNano())
+}
+
+// Next returns next random seed
+func (sg *SeedGenerator) Next() (*Seed, error) {
+	seed := Seed{}
+	_, err := rand.Read(seed[:])
```
golangcibot: G404: Use of weak random number generator (math/rand instead of crypto/rand)

ivanshibitov: fixed

functest/get_seed_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+package functest
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func askApi(t *testing.T) *getSeedResponse {
```
megge-dream: Name this func `getSeed` and move it to `utils_test.go`
Maybe its better to return `getSeedResponse.Seed`?

api/request_handler.go
```diff
@@ -301,3 +306,16 @@ func (rh *RequestHandler) ProcessIsAuthorized() (map[string]interface{}, error)
 
 	return result, nil
 }
+
+func (rh *RequestHandler) ProcessGetSeed() (map[string]interface{}, error) {
```
megge-dream: This func needs a comment too

PR #348, opened 2018-10-05
INS-512 consensus interface

AndreyBronin:
**- What I did**
Active node list consensus interface

### DIFF
--- a/network/consensus/consensus.go
+++ b/network/consensus/consensus.go
@@ -25,7 +25,6 @@ import (
 
 // Participant describes one consensus participant
 type Participant interface {
-	GetID() core.RecordRef
 	GetActiveNode() core.ActiveNode
 }
 


network/consensus/consensus.go
```diff
@@ -0,0 +1,65 @@
...
+type Consensus interface {
+	// DoConsensus is sync method, it make all consensus steps and returns boolean result
+	// method should be executed in goroutine
+	DoConsensus(ctx context.Context, self Participant, allParticipants []*Participant) (bool, error)
+}
+
+// Communicator interface is used to exchange messages between participants
+type Communicator interface {
+	// ExchangeData used in first consensus step to exchange data between participants
+	ExchangeData(ctx context.Context, p Participant, data []byte) ([]byte, error)
```
krestkrest: I think we should not make these interfaces fully abstract; we could use data structs and interfaces from core modules instead, making consensus modules separated from network layer only. We already have a RecordRef dependency, so it is natural to use other core modules too (especially keeping in mind that using interface{} is pain)

network/consensus/consensus.go
```diff
@@ -0,0 +1,57 @@
...
+	"context"
+	"errors"
+
+	"github.com/insolar/insolar/core"
+)
+
+// Participant describes one consensus participant
+type Participant interface {
+	GetID() core.RecordRef
+	GetActiveNode() core.ActiveNode
```
krestkrest: you can remove GetID because it returns GetActiveNode().NodeID

PR #363, opened 2018-10-10
INS-606: add per record ID locker

nordicdyno:
**- What I did**

Lock index writes.

Avoid local transaction errors on concurrent writes to the same indexes.


ledger/storage/transaction.go
```diff
@@ -40,9 +41,25 @@ func prefixkey(prefix byte, key []byte) []byte {
...
+	for _, id := range m.locks {
+		m.db.idlocker.Unlock(id)
+	}
+}
+
 // Commit tries to write transaction on disk. Returns error on fail.
 func (m *TransactionManager) Commit() error {
-	return m.txn.Commit(nil)
+	err := m.txn.Commit(nil)
+	m.releaseLocks()
```
nordicdyno: @andreyromancev probably locks should be released before Commit

nordicdyno: fixed in https://github.com/insolar/insolar/pull/363/commits/45bb695cdb50a03b3daa72bea8a90a9b1ff3353c

ledger/storage/lock.go
```diff
@@ -0,0 +1,95 @@
...
+	if mc, ok = l.m[*cid]; !ok {
+		mc = &mucount{Mutex: &sync.Mutex{}}
+		l.m[*cid] = mc
+	}
+	l.rwmu.Unlock()
+	return mc
+}
+
+// Lock locks mutex belonged to record ID.
+// If mutex not exists it has created in concurrent safe fashion.
```
andreyromancev: If mutex does not exist, it will be created in concurrent and safe fashion.

*`has created` means mutex itself created something, not was created. This line is confusing.

andreyromancev: Grammar nazi

PR #365, opened 2018-10-10
Ins 598 info api endpoint

ilyapt:
**- What I did**
- info method for bootstraper
- info route for api


api/info.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"net/http"
+
+	"github.com/insolar/insolar/core"
+	"github.com/insolar/insolar/log"
+	"github.com/pkg/errors"
+)
+
+func (ar *Runner) InfoHandler(c core.Components) func(http.ResponseWriter, *http.Request) {
```
LinnikD: we should comment public funcs

LinnikD: Does this handler really need to be Public?

PR #379, opened 2018-10-11
INS-576: use intermediate object state in validation

Vany:
**Now our validation stores object body and uses it from casebind, not last state from ledger**


core/vm.go
```diff
@@ -64,6 +64,7 @@ const (
 	caseRecordTypeUnexistent CaseRecordType = iota
 	CaseRecordTypeStart
 	CaseRecordTypeResult
+	CaseRecordTypeGetobject
```
ruz: s/Getobject/GetObject/

PR #381, opened 2018-10-12
Noissue consensus datarace fix

AndreyBronin:
**- What I did**
fix race bug fix

**- How I did it**
mutex

**- How to verify it**
tests with --race

### DIFF
--- a/network/consensus/base.go
+++ b/network/consensus/base.go
@@ -26,7 +26,7 @@ import (
 
 // exchangeResults is thread safe results struct
 type exchangeResults struct {
-	mutex *sync.RWMutex
+	mutex *sync.Mutex
 	data  map[core.RecordRef][]*core.ActiveNode
 	hash  []*NodeUnsyncHash
 }
@@ -54,7 +54,7 @@ func (r *exchangeResults) calculateResultHash() []*NodeUnsyncHash {
 
 func newExchangeResults(participantsCount int) *exchangeResults {
 	return &exchangeResults{
-		mutex: &sync.RWMutex{},
+		mutex: &sync.Mutex{},
 		data:  make(map[core.RecordRef][]*core.ActiveNode, participantsCount),
 		hash:  make([]*NodeUnsyncHash, participantsCount),
 	}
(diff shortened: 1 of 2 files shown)


network/consensus/base.go
```diff
@@ -24,28 +24,66 @@ import (
...
+	hash  []*NodeUnsyncHash
+}
+
+func (r *exchangeResults) writeResultData(id core.RecordRef, data []*core.ActiveNode) {
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+	r.data[id] = data
+}
+
+func (r *exchangeResults) getData(id core.RecordRef) []*core.ActiveNode {
```
golangcibot: func `(*exchangeResults).getData` is unused

network/consensus/base.go
```diff
@@ -24,28 +24,66 @@ import (
...
+
+func (r *exchangeResults) calculateResultHash() []*NodeUnsyncHash {
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+
+	for id, x := range r.data {
+		d, err := CalculateNodeUnsyncHash(id, x)
+		if err != nil {
+			log.Error(err)
+			continue
```
ivanshibitov: Is it ok to skip wrong calculation and continue working? May be method calculateResultHash should return error in that case?

network/consensus/base.go
```diff
@@ -24,28 +24,66 @@ import (
 	"github.com/insolar/insolar/log"
 )
 
-type exchangeResults map[core.RecordRef][]*core.ActiveNode
+// exchangeResults is thread safe results struct
+type exchangeResults struct {
+	mutex *sync.RWMutex
```
ivanshibitov: Why are you using RWMutex if you use only Lock/Unlock?

AndreyBronin: fixed

network/consensus/base.go
AndreyBronin: most likely there will never be an error. in consensus algorithm it's not very important if hash does not calculated

ivanshibitov: I mean, it looks like  r.data and r.hash should be the same length. If it is, and and one hash is not calculated - result may be surprising =)

PR #382, opened 2018-10-12
NOISSUE: Return correct error when fetching deactivated object.

andreyromancev:


core/reply/generic.go
```diff
@@ -0,0 +1,21 @@
...
+type Error struct {
+	ErrType ErrType
+}
+
+// Type implementation of Reply interface.
+func (e *Error) Type() core.ReplyType {
+	return TypeError
+}
+
+func (e *Error) Error() error {
```
nordicdyno: add godoc

PR #393, opened 2018-10-15
INS-587, INS-585: Remove authorization dummy

ooki2day:
**- What I did**

NetworkCoordinator integrated to service network; authorization on discovery node.

### DIFF
--- a/network/hostnetwork/dht.go
+++ b/network/hostnetwork/dht.go
@@ -63,7 +63,9 @@ type DHT struct {
 	nodeID            core.RecordRef
 	activeNodeKeeper  nodekeeper.NodeKeeper
 	majorityRule      int
+	uncheckedNodes    map[string][]byte
 	signChecker       func(msg core.Message) bool
+	signer            func(nonce []byte) []byte
 }
 
 // AuthInfo collects some information about authentication.
@@ -191,6 +193,8 @@ func NewDHT(
 	dht.auth.ReceivedKeys = make(map[string][]byte)
 
 	dht.subnet.SubnetIDs = make(map[string][]string)
+
+	dht.uncheckedNodes = make(map[string][]byte)
 
 	return dht, nil
 }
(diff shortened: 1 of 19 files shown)


network/hostnetwork/dht.go
```diff
@@ -1365,3 +1362,70 @@ func (dht *DHT) AddPossibleRelayID(id string) {
 func (dht *DHT) GetOriginHost() *host.Origin {
 	return dht.origin
 }
+
+func (dht *DHT) AddUncheckedNode(hostID id.ID, nonce []byte, ref core.RecordRef) {
+	unchecked := UncheckedNode{Ref: ref, Nonce: nonce}
+	dht.uncheckedNodes[hostID.String()] = unchecked
```
krestkrest: any time limit to store unchecked nodes? we should clean this map from time to time

network/hostnetwork/dht.go
```diff
@@ -1365,3 +1362,70 @@ func (dht *DHT) AddPossibleRelayID(id string) {
...
+	}
+	sign, err := ecdsa2.Sign(serialized, key)
+	if err != nil {
+		return errors.Wrap(err, "failed to sign a message")
+	}
+	msg.SetSign(sign)
+	return nil
+}
+
+func (dht *DHT) SignedNonceIsCorrect(hostID id.ID, signedNonce []byte) bool {
```
krestkrest: We could move all methods connected to sign checking to a separate interface, e.g. `SignChecker`, that would be accessible from `HostHandler` interface (via `GetSignChecker()` ). The implementation would have `privateKey` and `uncheckedNodes` as internal state. This step would make `HostHandler` interface lighter.

network/hostnetwork/packethandler_test.go
```diff
@@ -337,6 +366,14 @@ func (hh *mockHostHandler) SetSignChecker(func(msg core.Message) bool) {
 
 }
 
+func (hh *mockHostHandler) SetSigner(func(nonce []byte) []byte) {
```
AndreyBronin: define func param as type
type Signer func(nonce []byte) []byte

network/hostnetwork/dht.go
ooki2day: fix: added sign handler

PR #394, opened 2018-10-15
INS-623: recover on panics in insgorund

ruz:
don't crash whole daemon when a panic happens in a contract, return an
RPC error

### DIFF
--- a/logicrunner/goplugin/ginsider/ginsider.go
+++ b/logicrunner/goplugin/ginsider/ginsider.go
@@ -17,6 +17,7 @@
 package ginsider
 
 import (
+	"fmt"
 	"io/ioutil"
 	"net/rpc"
 	"os"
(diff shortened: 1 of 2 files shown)


logicrunner/goplugin/ginsider/ginsider.go
```diff
@@ -58,10 +59,21 @@ type RPC struct {
 	GI *GoInsider
 }
 
+func recoverRPC(err *error) {
+	if r := recover(); r != nil {
+		if err != nil && *err == nil {
+			*err = errors.New(fmt.Sprint(r))
```
ivanshibitov: May be is it better to join text of panic and text of error?
Because it looks like we loose text of panic if error exists, but print only text of panic

ruz: done

PR #408, opened 2018-10-17
INS-568 add checks for states and pulse's number

egorikas:
**- What I did**
Add checking for pulsar's state and pulse number. I think than the ifs allow us to be sure of impossibility of any damage to pulsar's state machine
**- How I did it**
Add ifs to the handler's methods
**- How to verify it**
Run tests

### DIFF
--- a/pulsar/pulsar_integration_test.go
+++ b/pulsar/pulsar_integration_test.go
@@ -146,12 +146,12 @@ func TestPulsar_SendPulseToNode(t *testing.T) {
 	assert.NoError(t, err)
 	count := 50
 	for (currentPulse == nil || currentPulse.PulseNumber == core.GenesisPulse.PulseNumber) && count > 0 {
-		time.Sleep(10 * time.Millisecond)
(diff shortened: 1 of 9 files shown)


pulsar/pulsar_integration_test.go
```diff
@@ -190,7 +190,7 @@ func TestTwoPulsars_Full_Consensus(t *testing.T) {
 				ConnectionType:      "tcp",
 				MainListenerAddress: ":1140",
 				BootstrapNodes:      []string{bootstrapAddress},
-				BootstrapListener:   configuration.Transport{Protocol: "UTP", Address: "127.0.0.1:18091", BehindNAT: false},
+				BootstrapListener:   configuration.Transport{Protocol: "UTP", Address: "127.0.0.1:1891", BehindNAT: false},
```
ita-sammann: Why port changed?

egorikas: Because sometimes there are some problems with erraddrinuse

PR #410, opened 2018-10-17
INS-657: Request recordref now passed from one call to another to cre...

Vany:
…ate new request based on previous

### DIFF
--- a/core/message/logicrunner.go
+++ b/core/message/logicrunner.go
@@ -47,27 +47,27 @@ type BaseLogicMessage struct {
 }
 
 // SetSign sets a signature to message.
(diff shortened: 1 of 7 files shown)


core/message/logicrunner.go
```diff
@@ -63,6 +65,11 @@ func (e *BaseLogicMessage) TargetRole() core.JetRole {
 	return core.RoleVirtualExecutor
 }
 
+// GetRequest returns RoleVirtualExecutor as routing target role.
+func (e *BaseLogicMessage) GetRequest() core.RecordRef {
```
golangcibot: receiver name e should be consistent with previous receiver name b for BaseLogicMessage

PR #431, opened 2018-10-19
Ins 584 trace in log

nordicdyno:
**- What I did**

add traceid to logger fields in context


testutils/recorder.go
```diff
@@ -0,0 +1,105 @@
+/*
```
golangcibot: package comment should not have leading space

PR #435, opened 2018-10-19
INS-672: added version to ActiveNode

krestkrest:
**- What I did**
Added version to ActiveNode


core/activenode.go
```diff
@@ -50,6 +50,8 @@ type ActiveNode struct {
 	PublicKey *ecdsa.PublicKey
 	// Addess is the network adress of the node
 	Address string
+	// Version of node software
+	Version string
```
AndreyBronin: May be will be better to store version as struct or json

AndreyBronin: May be will be better to store version as struct or json

PR #473, opened 2018-10-26
Ins 614 get state from bind

Vany:
**Now we are caching state on VE now**
also lots of bugs fixed


logicrunner/accessors.go
```diff
@@ -17,3 +17,83 @@
...
+	return oldObjectsRecords
+}
+
+func (lr *LogicRunner) addObjectCaseRecord(ref Ref, cr core.CaseRecord) {
+	lr.caseBindMutex.Lock()
+	lr.caseBind.Records[ref] = append(lr.caseBind.Records[ref], cr)
+	lr.caseBindMutex.Unlock()
+}
+
+func (lr *LogicRunner) lastObjectCaseRecord(ref Ref) core.CaseRecord {
```
golangcibot: func `(*LogicRunner).lastObjectCaseRecord` is unused

logicrunner/logicrunner_test.go
```diff
@@ -256,6 +257,7 @@ func New() (*Two, error) {
 
 func (r *Two) Hello(s string) (string, error) {
 	r.X ++
+//	return fmt.Sprintf("%d",r.X), nil
```
aalekseev-ru: can we remove this?

aalekseev-ru:
```suggestion

```

logicrunner/logicrunner_test.go
```diff
@@ -48,7 +48,7 @@ import (
 
 var icc = ""
 var runnerbin = ""
-var parallel = true
+var parallel = false
```
ruz: We shouldn't accept this change without explanation. What has happened with parallel tests that you had to disable them?

Vany: thanx, you have found it

logicrunner/logicrunner.go
```diff
@@ -232,127 +207,154 @@ func (lr *LogicRunner) Execute(ctx context.Context, inmsg core.SignedMessage) (c
...
+// make it private again when we start it serialize before sending
 type ObjectBody struct {
-	Object core.ObjectDescriptor
-	Class  core.ClassDescriptor
-	Code   core.CodeDescriptor
+	objDescriptor   core.ObjectDescriptor
+	Object          []byte
+	ClassHeadRef    *Ref
+	CodeMachineType core.MachineType
+	CodeRef         *Ref
```
ruz: I don't say this should be reverted, but we should aim different direction. I believe we should use more of Descriptors objects in our code and make sure they are passive carriers of data with declared interfaces. As we move forward towards new features we need more and more fields from each structure. My opinion is quite strong that ObjectBody should go away as a struct and instead we should keep separate fields for object, class and code. The last two can be pointers as we can be executing multiple objects with one class and thus the code, so we can share descriptors and minimize net travels for ledger accesses.

Do you agree?

logicrunner/logicrunner.go
```diff
@@ -232,127 +207,154 @@ func (lr *LogicRunner) Execute(ctx context.Context, inmsg core.SignedMessage) (c
...
-}
-
+// ObjectBody is an inner representation of object and all it accessory
+// make it private again when we start it serialize before sending
 type ObjectBody struct {
-	Object core.ObjectDescriptor
-	Class  core.ClassDescriptor
-	Code   core.CodeDescriptor
+	objDescriptor   core.ObjectDescriptor
+	Object          []byte
```
ruz: Object's memory is duplicated as Object field and objDescriptor.Memory() method. This is wrong.

logicrunner/logicrunner.go
Vany: Main problem is we can't transfer it to another machine and reconstruct it there, so we need something transferable.

logicrunner/logicrunner.go
Vany: no, it's a link.
but you can not get object on one machine and update it on another, it is a workaround for ledger

PR #479, opened 2018-10-29
Noissue add jq to docker

ivanshibitov:
**- What I did**
Add jq and other soft to docker


application/contract/wallet/wallet.go
```diff
@@ -99,6 +100,7 @@ func (w *Wallet) Accept(aRef *core.RecordRef) error {
 	return nil
 }
 
+// GetTotalBalance get total balance
```
megge-dream: _GetTotalBalance get**s**_

PR #484, opened 2018-10-30
NET-4: new transport

krestkrest:
**- What I did**
Implemented new network on transport layer to replace old DHT transport
**- How to verify it**
run tests


network/hostnetwork/transport.go
```diff
@@ -0,0 +1,256 @@
...
+		select {
+		case msg := <-h.transport.Packets():
+			if msg == nil {
+				log.Error("HostNetwork receiving channel is closed")
+				break
+			}
+			if msg.Error != nil {
+				log.Warnf("Received error response: %s", msg.Error.Error())
+			}
+			h.processMessage(msg)
```
dmitry-panchenko: We should process packet in another goroutine.

network/hostnetwork/transport.go
```diff
@@ -0,0 +1,256 @@
...
+		return errors.New("double listen initiated")
+	}
+	go h.listen()
+	return h.transport.Start()
+}
+
+func (h *hostTransport) listen() {
+	for {
+		select {
+		case msg := <-h.transport.Packets():
```
dmitry-panchenko: Why `msg`? Could we name it `packet`?

network/hostnetwork/transport.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+}
+
+func (f future) GetRequest() network.Request {
+	request := transport.Future(f).Request()
+	return (*packetWrapper)(request)
+}
+
+// Listen start listening to network requests, should be started in goroutine.
+func (h *hostTransport) Listen() error {
```
dmitry-panchenko: `.Listen()` and `.Disconnect()` methods are very similar to Component's `.Start()` and `.Stop()` methods.
Could we transform transport to Component?

network/transport/transport.go
```diff
@@ -68,6 +68,7 @@ func NewTransport(cfg configuration.Transport, proxy relay.Proxy) (Transport, er
 	case "KCP":
 		return newKCPTransport(conn, proxy, publicAddress)
 	default:
+		closePlease(conn)
```
dmitry-panchenko: `closeVerbose`?

network/hostnetwork/transport.go
krestkrest: I wanted to distinguish it from package `packet`

network/interfaces.go
```diff
@@ -60,17 +61,19 @@ type HostNetwork interface {
 	PublicAddress() string
 
 	// SendRequest send request to a remote node.
-	SendRequest(Request) (Future, error)
+	SendRequest(request Request, receiver core.RecordRef) (Future, error)
```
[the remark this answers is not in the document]
krestkrest: argument names were added to specify that the second argument is the receiver's node identifier

PR #485, opened 2018-10-30
Ins 728 node domain with auth

ilyapt:
**- What I did**
- registerNode called with sign


functest/register_node_test.go
```diff
@@ -17,30 +17,42 @@
...
+}
 
-	response := &registerNodeResponse{}
-	unmarshalResponseWithError(t, body, response)
+func registerNodeSignedCall(params ...interface{}) (*registerAnswer, error) {
+	res, err := signedRequest(&root, "RegisterNode", params...)
+	if err != nil {
+		return nil, err
+	}
+	fmt.Println(res.(string))
```
ivanshibitov: Do we need it?

PR #486, opened 2018-10-30
INS-722: Expiry queue message bus

pavlushev:
Created a queue with expiration:

- the queue is created during the creation of the MessageBus
- In the "Send" method, we add msg to the queue; if the queue is unreadable, messages will be automatically deleted


messagebus/expiryqueue.go
```diff
@@ -0,0 +1,109 @@
...
+		if eq.items.Len() > 0 {
+
+			eq.mutex.Lock()
+			for {
+				e := eq.items.Front()
+				if e == nil {
+					break
+				}
+				elem := e.Value.(*Item)
+				sleepTime = elem.expireAt.Sub(time.Now())
```
golangcibot: should use time.Until instead of t.Sub(time.Now())

pavlushev: fixed

PR #488, opened 2018-10-31
INS-692: new return nil

megge-dream:
closes INS-692

**- What I did**
Add check if nil in sc constructor wrapper.

**- How to verify it**
Run tests.

### DIFF
--- a/logicrunner/logicrunner_test.go
+++ b/logicrunner/logicrunner_test.go
@@ -1462,7 +1462,7 @@ func (r *Two) Hello() (*string, error) {
 		Arguments: goplugintestutils.CBORMarshal(t, []interface{}{}),
 	}
 	key, _ := cryptoHelper.GeneratePrivateKey()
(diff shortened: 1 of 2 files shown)


logicrunner/logicrunner_test.go
```diff
@@ -1390,3 +1390,90 @@ func New(n int) (*Child, error) {
...
+	"github.com/insolar/insolar/logicrunner/goplugin/foundation"
+)
+
+type Two struct {
+	foundation.BaseContract
+}
+func New() (*Two, error) {
+	return nil, nil
+}
+func (r *Two) Hello() (*string, error) {
```
Vany: Do we actually use this method ?

megge-dream: Otherwise we can't build contract. Add comment about it.

Vany: please crete an issue about it

PR #489, opened 2018-10-31
NOISSUE: fix flaky ledger test

nordicdyno:
NOISSUE: fix flaky ledger test
'after' is not reliable because lack of synchronization


egorikas:
It's my school :)

PR #490, opened 2018-10-31
Net 35 udp transport interface

ivanshibitov:
**- What I did**
1) reuse base transport interface for new udp transport. Now it uses common packet format
2) add logic for using different serialization function
3) add tests
4) skip TestSendBigPacket since it doesn't work for ump transport

**- How to verify it**
Run tests


network/transport/udp.go
```diff
@@ -0,0 +1,105 @@
...
+		go udpT.handleAcceptedConnection(buf[:n], addr)
+	}
+}
+
+// Stop stops networking.
+func (udpT *udpTransport) Stop() {
+	udpT.mutex.Lock()
+	defer udpT.mutex.Unlock()
+
+	log.Info("Stop UTP transport")
```
krestkrest: Stop UDP transport

network/transport/udp.go
```diff
@@ -0,0 +1,105 @@
...
+	log.Debug("udpTransport.send: len = ", len(data))
+	_, err = udpConn.Write(data)
+	return errors.Wrap(err, "Failed to write data")
+}
+
+// Start starts networking.
+func (udpT *udpTransport) Start() error {
+	log.Info("Start UDP transport")
+	for {
+		buf := make([]byte, 1400)
```
krestkrest: move MTU to constants (or even configurable options) and return error when we try to send packet with size more than MTU

ivanshibitov: fixed

network/transport/udp.go
ivanshibitov: fixed

network/transport/base.go
```diff
@@ -40,9 +40,10 @@ type baseTransport struct {
 	mutex   *sync.RWMutex
 	futures map[packet.RequestID]Future
 
-	proxy         relay.Proxy
-	publicAddress string
-	sendFunc      func(recvAddress string, data []byte) error
+	proxy          relay.Proxy
+	publicAddress  string
+	sendFunc       func(recvAddress string, data []byte) error
```
LinnikD: This can be done using Interface transport. Should we fix this now?

PR #513, opened 2018-11-02
NET-40, Net-41 phase1 exchange interface

AndreyBronin:
**- What I did**
Consensus communicator interface proposal
Using DI in phases package
initial tests


testutils/network/node_network_mock.go
```diff
@@ -0,0 +1,398 @@
+package network
```
krestkrest: we need go generate rule for this in one of other source files

PR #527, opened 2018-11-07
INS-686 Parent for app logic

aalekseev-ru:
**- What I did**
add Parent to CallContext
**- How I did it**
take from ledger, pass to es.CallContext
**- How to verify it**
run tests


core/record.go
```diff
@@ -103,6 +103,14 @@ func (ref RecordRef) String() string {
 	return base58.Encode(ref[:])
 }
 
+// After CBOR Marshal/Unmarshal Ref can be converted to byte slice, this converts it back
```
Vany:
```suggestion
// FromSlice : After CBOR Marshal/Unmarshal Ref can be converted to byte slice, this converts it back
```

PR #543, opened 2018-11-08
INS-747: cbor for all

megge-dream:
closes INS-747

**- What I did**
Replace different usage of cbor with one function.

**- How to verify it**
Run tests.


logicrunner/casebind.go
```diff
@@ -28,14 +28,11 @@ import (
 	"github.com/insolar/insolar/core/message"
 	"github.com/insolar/insolar/core/reply"
 	"github.com/pkg/errors"
-	"github.com/ugorji/go/codec"
 	"golang.org/x/crypto/sha3"
 )
 
 func HashInterface(in interface{}) []byte {
```
Vany: this is a wrong function now and must be removed, please change all it's usages for core function.

PR #544, opened 2018-11-08
Net 75 construct packet for phase2

ivanshibitov:
**- What I did**
Implement PacketRoutable interface for Phase1 & 2 packets 

**- How to verify it**
Run tests


consensus/phases/data.go
```diff
@@ -161,7 +196,7 @@ type NodeJoinClaim struct {
 	JoinsAfter              uint32
 	NodeRoleRecID           uint32
 	NodeRef                 core.RecordRef
-	//NodePK                  ecdsa.PrivateKey // WTF: should it be Public key?
+	NodePK                  ecdsa.PublicKey
```
krestkrest: we should not use specific crypto protocols here (use byte slice instead)

consensus/phases/data.go
```diff
@@ -219,3 +254,49 @@ type DeviantBitSet struct {
...
+
+func (p2p *Phase2Packet) GetPacketHeader() (*RoutingHeader, error) {
+	header := &RoutingHeader{}
+
+	if p2p.packetHeader.PacketT != NetworkConsistency {
+		return nil, errors.New("Phase2Packet.GetPacketHeader: wrong packet type")
+	}
+
+	header.PacketType = types.Phase2
+	header.OriginID = p2p.packetHeader.OriginNodeID
```
LinnikD: p2p name can confuse. Maybe better FirstPhasePacket, SecondPhasePacket and fpp spp for recievers

PR #552, opened 2018-11-09
use new DI in logic runner, first step

ruz:


logicrunner/logicrunner.go
```diff
@@ -59,15 +59,17 @@ func (es *ExecutionState) Unlock() {
 
 // LogicRunner is a general interface of contract executor
 type LogicRunner struct {
-	Executors       [core.MachineTypesLastID]core.MachineLogicExecutor
+	MessageBus core.MessageBus `inject:""`
+	Ledger     core.Ledger     `inject:""`
+	Network    core.Network    `inject:""`
+
 	ArtifactManager core.ArtifactManager
```
aalekseev-ru: ArtifactManager isn't a component? Or just not yet?

ruz: It's not a component. It's part of ledger.

PR #556, opened 2018-11-09
Net 109 implementation extract packet

ivanshibitov:
**- What I did**
Implement ExtractPacket

**- How to verify it**
Run tests

### DIFF
--- a/consensus/packets/interfaces.go
+++ b/consensus/packets/interfaces.go
@@ -17,8 +17,10 @@
 package packets
 
 import (
-	"errors"
 	"io"
+	"strconv"
+
+	"github.com/pkg/errors"
 
 	"github.com/insolar/insolar/core"
 	"github.com/insolar/insolar/network/transport/packet/types"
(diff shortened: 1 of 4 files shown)


consensus/packets/interfaces.go
```diff
@@ -48,5 +50,31 @@ type ConsensusPacket interface {
...
+	err := header.Deserialize(reader)
+	if err != nil {
+		return nil, errors.New("[ ExtractPacket ] Can't read packet header")
+	}
+
+	var packet ConsensusPacket
+	switch header.PacketT {
+	case Phase1:
+		packet1 := &Phase1Packet{}
+		err := packet1.DeserializeWithoutHeader(reader, &header)
```
krestkrest: move `DeserializeWithoutHeader` to `Serializer` interface and remove copy-paste lines 63-66, 70-74

ivanshibitov: It means that all structures will have to realize it. It's better make one more separate interface with only this function

PR #562, opened 2018-11-12
INS-737 build proxy without methods

aalekseev-ru:
**- What I did**
Now you can build contracts without methods.
**- How I did it**
If there is no methods except constructors we doesn't import foundation in proxy.
**- How to verify it**
I change broken test and now it works without hacks


logicrunner/goplugin/preprocessor/main.go
```diff
@@ -361,6 +364,12 @@ func (pf *ParsedFile) generateImports(wrapper bool) map[string]bool {
 			extendImportsMap(pf, fun.Type.Results, imports)
 		}
 	}
+
+	if imports[fmt.Sprintf(`"%s"`, foundationPath)] && removeFoundation {
+		delete(imports, fmt.Sprintf(`"%s"`, foundationPath))
+		fmt.Println("WARNING: build proxy without methods")
```
Vany: please move it closer to methods parsing. 
We do not know anything about methods here

PR #569, opened 2018-11-13
INS-733 new ES Error Wrapper

aalekseev-ru:
**- What I did**
added es.ErrorWrapper
**- How I did it**
add new method and new data for it, change old tests
**- How to verify it**
run tests

### DIFF
--- a/logicrunner/logicrunner.go
+++ b/logicrunner/logicrunner.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"encoding/gob"
 	"net"
+	"strings"
 	"sync"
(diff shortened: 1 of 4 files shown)


logicrunner/logicrunner.go
```diff
@@ -53,6 +56,31 @@ type ExecutionState struct {
 	objectbody *ObjectBody
 }
 
+type LogicRunnerError struct {
```
golangcibot: type name will be used as logicrunner.LogicRunnerError by other packages, and that stutters; consider calling this Error

PR #584, opened 2018-11-15
Ins 764 start validation chain

Vany:
**Now we can use special annotation on methods to allow call it from api (Without initial caller and request)**

### DIFF
--- a/logicrunner/goplugin/ginsider/ginsider.go
+++ b/logicrunner/goplugin/ginsider/ginsider.go
@@ -90,6 +90,23 @@ func (t *RPC) CallMethod(args rpctypes.DownCallMethodReq, reply *rpctypes.DownCa
 		return errors.Wrapf(err, "Couldn't get plugin by code reference %s", args.Code.String())
 	}
 
+	if args.Context.Caller.Equal(core.RecordRef{}) {
+		attr, err := p.Lookup("INSATTR_" + args.Method + "_API")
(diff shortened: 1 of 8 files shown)


logicrunner/logicrunner.go
```diff
@@ -36,12 +36,14 @@ import (
 	"github.com/insolar/insolar/logicrunner/goplugin"
 )
 
+const deny_test_call_constructor = false
```
golangcibot: don't use underscores in Go names; const deny_test_call_constructor should be denyTestCallConstructor

LinnikD: +

logicrunner/goplugin/ginsider/ginsider.go
```diff
@@ -90,6 +90,23 @@ func (t *RPC) CallMethod(args rpctypes.DownCallMethodReq, reply *rpctypes.DownCa
...
+				err, "Calling non INSATTRAPI method %s (code ref: %s)",
+				args.Method, args.Code.String(),
+			)
+		}
+		api, ok := attr.(*bool)
+		if !ok {
+			return errors.Errorf("INSATTRAPI attribute for method %s is not boolean", args.Method)
+		}
+		if !*api {
+			return errors.Errorf("Calling non INSATTRAPI method ")
```
aalekseev-ru: maybe add here ref too?

PR #593, opened 2018-11-16
NET-116: Interface fsm

LinnikD:
**- What I did**
 Interface fsm

### DIFF
--- a/network/state/switcher.go
+++ b/network/state/switcher.go
@@ -16,7 +16,9 @@
 
 package state
 
-type State int
+import (
+	"github.com/insolar/insolar/network"
+)
 
 const (
 	// NoNetwork state means that nodes doesn`t match majority_rule
@@ -31,10 +33,6 @@ const (
 	CompleteNetwork
 )
 
-type NetworkSwitcher interface {
-	GetState() State
-}
-
 type NetworkSwitcherImpl struct {
 }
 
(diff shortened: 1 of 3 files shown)


network/state/switcher.go
```diff
@@ -0,0 +1,45 @@
...
+
+package state
+
+import (
+	"github.com/insolar/insolar/network"
+)
+
+const (
+	// NoNetwork state means that nodes doesn`t match majority_rule
+	NoNetwork = iota
```
AndreyBronin: use go:generate stringer for sexy logging

network/state/switcher.go
```diff
@@ -0,0 +1,45 @@
...
+	VoidNetwork
+	// JetlessNetwork state means that every Jet need proof completeness of stored data
+	JetlessNetwork
+	// AuthorizationNetwork state means that every node need to validate ActiveNodeList using NodeDomain
+	AuthorizationNetwork
+	// CompleteNetwork state means network is ok and ready for proper work
+	CompleteNetwork
+)
+
+type NetworkSwitcherImpl struct {
```
AndreyBronin: impl should not be exported

network/state/switcher.go
LinnikD: It is not necessary for now

PR #610, opened 2018-11-21
NET-163: nodekeeper fix

krestkrest:
**- What I did**
Fixed bug that caused to nodes duplication in nodekeeper role index
**- How to verify it**
run tests


network/nodenetwork/recordref_set.go
```diff
@@ -0,0 +1,54 @@
...
+
+func (s *recordRefSet) Contains(ref core.RecordRef) bool {
+	_, ok := s.data[ref]
+	return ok
+}
+
+func (s *recordRefSet) Collect() []core.RecordRef {
+	result := make([]core.RecordRef, len(s.data))
+	i := 0
+	for ref, _ := range s.data {
```
golangcibot: should omit 2nd value from range; this loop is equivalent to `for ref := range ...`

PR #616, opened 2018-11-21
Ins-792: replicate all records

nordicdyno:
**- What I did**

Add blob and jetdrop key values to heavy replication


ledger/pulsemanager/heavy_test.go
```diff
@@ -172,7 +172,7 @@ func TestPulseManager_SendToHeavy(t *testing.T) {
 }
 
 func setpulse(ctx context.Context, pm core.PulseManager, pulsenum int) error {
-	fmt.Println("CALL PulseManager.Set with pulse", pulsenum)
+	// fmt.Println("CALL PulseManager.Set with pulse", pulsenum)
```
egorikas: Should we delete it?

PR #646, opened 2018-11-26
NET-136: json rpc routing

megge-dream:
closes NET-136

**- What I did**
Use JSON-RPC in API component.

**- How to verify it**
Run tests.


api/call.go
```diff
@@ -33,7 +33,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-type request struct {
+type APIRequest struct {
```
golangcibot: type name will be used as api.APIRequest by other packages, and that stutters; consider calling this Request (from `golint`)

PR #647, opened 2018-11-26
simplify how we deal with lock during execution

ruz:
* use closures with local variable rather than state struct,
  local variable makes it clear that this thing is not used elsewhere
  and makes problem isolated
* make only one defer conditional the one in the beginning, less logic
  -> less chances of error

### DIFF
--- a/logicrunner/logicrunner.go
+++ b/logicrunner/logicrunner.go
@@ -44,7 +44,6 @@ type ExecutionState struct {
 	Ref    *Ref
 	Method string
 
-	noWait      bool
 	validate    bool
 	insContext  context.Context
 	callContext *core.LogicCallContext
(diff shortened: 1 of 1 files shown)


logicrunner/logicrunner.go
```diff
@@ -443,9 +442,9 @@ func (lr *LogicRunner) getObjectMessage(es *ExecutionState, objref Ref) error {
 func (lr *LogicRunner) executeMethodCall(es *ExecutionState, m *message.CallMethod, vb ValidationBehaviour) (core.Reply, error) {
 	ctx := es.insContext
 
-	es.noWait = false
+	delayedUnlock = false
```
golangcibot: undeclared name: delayedUnlock (from `typecheck`)

PR #653, opened 2018-11-26
Net 173 tcp transport

ivanshibitov:
**- What I did**
Add tcp transport

**- How to verify it**
Run tests

### DIFF
--- a/network/transport/tcp.go
+++ b/network/transport/tcp.go
@@ -1,0 +1,116 @@
+/*
+ *    Copyright 2018 Insolar
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


network/transport/tcp.go
```diff
@@ -0,0 +1,116 @@
...
+
+	transport.sendFunc = transport.send
+
+	return transport, nil
+}
+
+func (tcp *tcpTransport) send(recvAddress string, data []byte) error {
+
+	// TODO: may be try to send second time if error
+	// TODO: skip resolving every time by caching result
```
dmitry-panchenko: You do not need this TODO

ivanshibitov: done

PR #656, opened 2018-11-27
Net 139 fsm dummy

LinnikD:
**- What I did**
v1 Net FSM implementation


network/nodenetwork/nodekeeper.go
```diff
@@ -112,6 +112,10 @@ type nodekeeper struct {
 	syncLock sync.Mutex
 }
 
+func (nk *nodekeeper) IsBootstrapped() bool {
```
megge-dream: Add comment for this func

LinnikD: done

network/state/switcher.go
```diff
@@ -17,19 +17,46 @@
...
+
+// OnPulse method checks current state and finds out reasons to update this state
+func (ns *NetworkSwitcher) OnPulse(ctx context.Context, pulse core.Pulse) error {
+	inslogger.FromContext(ctx).Info("Current NetworkSwitcher state is: %s", ns.state)
+
+	if ns.SwitcherWorkAround.IsBootstrapped() {
+		ns.stateLock.Lock()
+		ns.state = core.CompleteNetworkState
+		ns.stateLock.Unlock()
+		inslogger.FromContext(ctx).Info("Current NetworkSwitcher state switched to: %s", ns.state)
```
ivanshibitov: Concurrent access to ns.state
Save ns.state to another variable inside Lock and then use it here

ivanshibitov: Or even print it inside lock

PR #661, opened 2018-11-28
NET-176: apirequester

megge-dream:
closes NET-176

**- What I did**
apirequester makes several requests to API and return results of them.

**- How to verify it**
Run tests.

### DIFF
--- a/cmd/apirequester/main.go
+++ b/cmd/apirequester/main.go
@@ -83,9 +83,20 @@ func main() {
 	check("can not set 'error' level on logger:", err)
 	rootMember = getRootMemberInfo(rootmemberkeys)
 
+	// you can modify this manual tests by commenting any of this functions or/and add some new functions if necessary
+
+	// make one request to create new member
 	oneSimpleRequest()
+
+	// make several (10) requests to create new member (every request make call to RootMember instance)
 	severalSimpleRequestToRootMember()
+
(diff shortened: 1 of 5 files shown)


cmd/apirequester/main.go
```diff
@@ -0,0 +1,90 @@
...
+	err = json.Unmarshal(rawConf, &keys)
+	check("problem with unmarshaling root member keys", err)
+
+	return memberInfo{getRootMemberRef(), keys.Private}
+}
+
+func main() {
+	parseInputParams()
+
+	log.SetLevel("error")
```
golangcibot: Error return value of `log.SetLevel` is not checked (from `errcheck`)

cmd/apirequester/main.go
```diff
@@ -0,0 +1,91 @@
...
+	return memberInfo{getRootMemberRef(), keys.Private}
+}
+
+func main() {
+	parseInputParams()
+
+	err := log.SetLevel("error")
+	check("can not set 'error' level on logger:", err)
+	rootMember = getRootMemberInfo(rootmemberkeys)
+
```
LinnikD: leave here some comments or add it to readme.
Tell about currently exists functions and that you can modify this manual tests by commenting this call and add some new functions if necessary

megge-dream: done

PR #663, opened 2018-11-28
Net 133 register node call

ilyapt:
**- What I did**
- RegisterNode() checks who call it
- Fixed register_node_tests


application/contract/rootdomain/rootdomain.go
```diff
@@ -53,6 +53,10 @@ func (rd *RootDomain) CreateMember(name string, key string) (string, error) {
 	return m.GetReference().String(), nil
 }
 
+func (rd *RootDomain) GetRootMemberRef() (*core.RecordRef, error) {
```
megge-dream: Add comment for this func

PR #671, opened 2018-11-28
Noissue functest separate build

AndreyBronin:
**- What I did**
functest build separated

### DIFF
--- a/functest/common_test.go
+++ b/functest/common_test.go
@@ -1,3 +1,5 @@
+// +build functest
+
 /*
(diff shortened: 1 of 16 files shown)


functest/common_test.go
```diff
@@ -1,3 +1,5 @@
+// +build functest
```
krestkrest: should we have copyright on the top?

AndreyBronin: no, coz +build tag comment must be first line

PR #705, opened 2018-12-03
[WIP] INS-122: register node

ilyapt:
**- What I did**

**- How I did it**

**- How to verify it**


core/message/node_sign.go
```diff
@@ -0,0 +1,44 @@
+/*
```
LinnikD: node_sign -> nodesign here and in others

PR #709, opened 2018-12-04
NET-8 NET-29 testcase node connect

AndreyBronin:
**- What I did**
integration test -  node join
NodeNetwork and PhaseManager wrappers for test usage

**- How I did it**

**- How to verify it**
nope

### DIFF
--- a/network/servicenetwork/integr_test.go
+++ b/network/servicenetwork/integr_test.go
@@ -176,7 +176,7 @@ func (s *testSuite) createNetworkNode(t *testing.T) networkNode {
 }
 
 func (s *testSuite) TestNodeConnect() {
-	//s.T().Skip("fixme")
+	s.T().Skip("fix nodes auth, generate valid certs!")
 
 	phasesResult := make(chan error)
 	s.InitNodes()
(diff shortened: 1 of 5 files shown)


network/servicenetwork/wrappers_test.go
```diff
@@ -0,0 +1,129 @@
...
+import (
+	"context"
+
+	consensus "github.com/insolar/insolar/consensus/packets"
+	"github.com/insolar/insolar/consensus/phases"
+	"github.com/insolar/insolar/core"
+	"github.com/insolar/insolar/network"
+)
+
+type nodeKeeperWrapper struct {
```
krestkrest: can we use minimock instead of this boilerplate wrapper?

network/servicenetwork/integr_test.go
```diff
@@ -156,14 +176,23 @@ func (s *testSuite) createNetworkNode(t *testing.T) networkNode {
 }
 
 func (s *testSuite) TestNodeConnect() {
```
krestkrest: This test only covers case when discovery nodes connect to each other and form a new network. Do we plan to write a testcase in this task where a joiner node is connecting to an already formed network of discovery nodes?

PR #715, opened 2018-12-04
Ins 133 cert config

LinnikD:
**- What I did**

No cert and conf interception anymore


pulsar/pulsar_integration_test.go
```diff
@@ -183,7 +186,7 @@ func initNetwork(ctx context.Context, t *testing.T, bootstrapHosts []string) (*l
 	return tempLedger, cleaner, serviceNetwork, address
 }
 
-func TestPulsar_SendPulseToNode(t *testing.T) {
+func _TestPulsar_SendPulseToNode(t *testing.T) {
```
AndreyBronin: Use t.Skip("reason") instead

LinnikD: already unskipped, but thx!

PR #716, opened 2018-12-05
NOISSUE: add --keep-temp flag to insgocc

afiskon:

The support of --keep-temp flag was added to insgocc utility. When user provides this flag, insgocc keeps temporary files created during `insgocc compile`, which includes files generated from templates (.go.tpl). This is useful during debugging.


Vany:
Great!

PR #725, opened 2018-12-05
INS-873: node registration test

ilyapt:
**- What I did**
- wrote functional test for node registration
- fixed serialization and signing for node certificates

**- How to verify it**
- run functests


certificate/certificate.go
```diff
@@ -141,9 +141,9 @@ func (cert *Certificate) serializeNetworkPart() []byte {
 }
 
 // SignNetworkPart signs network part in certificate
-func (cert *Certificate) SignNetworkPart(key crypto.PrivateKey) ([]byte, error) {
+func (cert *Certificate) SignNetworkPart(key crypto.PrivateKey, serialized []byte) ([]byte, error) {
```
LinnikD: No need to pass serialized. We need to call serialize() inside SignNetworkPart method

PR #755, opened 2018-12-10
INS-919: serialization codegen (part 1 of 2, without recursive enumeration of all types used in the contract)

afiskon:
**- What I did**

Proposed patch speeds up the serialization and deserialization of the contracts by using code generation feature of `codec` library. The patch doesn't implement recursive code generation for all types used in the contract since this is a much more complicated feature and it deserves a separate code review. 

**- How to verify it**

1. Build the system (e.g. `make build`) from branch `master`
2. Execute `./bin/insgocc compile --keep-temp ./application/contract/wallet/wallet.go`
3. Remember the size of wallet.so and the content of the temp directory (the path is reported by insgocc utility). Repeat for other contracts.
4. Apply the path
5. Execute `make build` once again and repeat step 2
6. Make sure the resulting .so files are much larger, there are extra files in the temp directory, and the temp directory is in `$GOPATH` now.

Speed-up of serialization and deserialization of the contracts.

### DIFF
--- a/cmd/insgocc/insgocc.go
+++ b/cmd/insgocc/insgocc.go
@@ -18,16 +18,16 @@ package main
 
 import (
 	"fmt"
+	"github.com/insolar/insolar/logicrunner/goplugin/preprocessor"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"go/build"
 	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
 	"path"
 	"path/filepath"
-
-	"github.com/insolar/insolar/logicrunner/goplugin/preprocessor"
-	"github.com/pkg/errors"
-	"github.com/spf13/cobra"
 )
 
 type outputFlag struct {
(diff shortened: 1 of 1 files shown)


cmd/insgocc/insgocc.go
```diff
@@ -235,13 +250,62 @@ func main() {
...
+			}
+
+			// 2. Execute `codecgen` for all .go files in the directory
+			files, err := filepath.Glob("./*.go")
+			if err != nil {
+				fmt.Printf("Unable to enumerate files in directory %s: %v\n", tmpDir, err)
+				os.Exit(1)
+			}
+
+			expr, _ := regexp.Compile("^(.*)\\.go$")
```
golangcibot: should use raw string (`...`) with regexp.Compile to avoid having to escape twice (from `megacheck`)

afiskon:
@ruz thanks for bringing it to our attention. Ivan from applogic team says the are using `insolard genesis` command for generating contracts. I will investigate what it does and modify the PR accordingly.

afiskon:
@ruz you are right, apparently it's done in `func (cb *ContractsBuilder) Build(contracts map[string]string) error` despite the fact that this method is placed in goplugin**testutils** package.

Considering this, and also the fact that according to the benchmark CPU is not a bottleneck on 3-node test, I would like to close this PR.

PR #783, opened 2018-12-12
INS-931: lock mb before complete network

megge-dream:
closes INS-931

**- What I did**
Lock MessageBus if network state is not "CompleteNetworkState"

**- How to verify it**
Run tests

### DIFF
--- a/messagebus/messagebus.go
+++ b/messagebus/messagebus.go
@@ -56,10 +56,12 @@ type MessageBus struct {
 // NewMessageBus creates plain MessageBus instance. It can be used to create Player and Recorder instances that
 // wrap it, providing additional functionality.
 func NewMessageBus(config configuration.Configuration) (*MessageBus, error) {
-	return &MessageBus{
+	mb := &MessageBus{
 		handlers:     map[core.MessageType]core.MessageHandler{},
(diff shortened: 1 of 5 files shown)


messagebus/messagebus.go
```diff
@@ -189,7 +191,7 @@ func (mb *MessageBus) SendParcel(
 		return nil, err
 	}
 
-	scope.Unlock()
+	scope.Unlock(ctx, "Sending parcel done")
```
LinnikD: Why do you need double unlock here and in defer?

network/state/switcher.go
```diff
@@ -56,10 +86,62 @@ func (ns *NetworkSwitcher) OnPulse(ctx context.Context, pulse core.Pulse) error
...
 		inslogger.FromContext(ctx).Info("Current NetworkSwitcher state switched to: %s", ns.state)
 	}
 
 	return nil
 }
+
+func (ns *NetworkSwitcher) AcquireGlobalLock(ctx context.Context, caller string) {
+	ns.mbLocksLock.Lock()
+
+	callerLock, ok := ns.mbLocks[caller]
```
golangcibot: this value of `callerLock` is never used (from `megacheck`)

PR #786, opened 2018-12-13
INS-940: code caching in artifact manager

ilyapt:
**- What I did**
- added code cache to artefact manager's GetCode()
- removed descriptors cache
- wrote unit test for GetCode()

**- How to verify it**
- run artefact manager's unit test

### DIFF
--- a/ledger/artifactmanager/artifactmanager.go
+++ b/ledger/artifactmanager/artifactmanager.go
@@ -40,8 +40,15 @@ type LedgerArtifactManager struct {
 	db                         *storage.DB
 	DefaultBus                 core.MessageBus                 `inject:""`
 	PlatformCryptographyScheme core.PlatformCryptographyScheme `inject:""`
+	cacheLock                  *sync.Mutex
+	cache                      map[core.RecordRef]cacheEntry
 
 	getChildrenChunkSize int
+}
+
+type cacheEntry struct {
+	sync.Mutex
+	desc core.CodeDescriptor
 }
 
 // State returns hash state for artifact manager.
@@ -52,7 +59,12 @@ func (m *LedgerArtifactManager) State() ([]byte, error) {
 
 // NewArtifactManger creates new manager instance.
 func NewArtifactManger(db *storage.DB) *LedgerArtifactManager {
-	return &LedgerArtifactManager{db: db, getChildrenChunkSize: getChildrenChunkSize}
+	return &LedgerArtifactManager{
+		db:                   db,
+		getChildrenChunkSize: getChildrenChunkSize,
+		cacheLock:            &sync.Mutex{},
+		cache:                make(map[core.RecordRef]cacheEntry, 0),
+	}
 }
 
 // GenesisRef returns the root record reference.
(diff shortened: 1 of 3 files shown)


ledger/artifactmanager/artifactmanager.go
```diff
@@ -52,7 +59,12 @@ func (m *LedgerArtifactManager) State() ([]byte, error) {
 
 // NewArtifactManger creates new manager instance.
 func NewArtifactManger(db *storage.DB) *LedgerArtifactManager {
-	return &LedgerArtifactManager{db: db, getChildrenChunkSize: getChildrenChunkSize}
+	return &LedgerArtifactManager{
+		db:                   db,
+		getChildrenChunkSize: getChildrenChunkSize,
+		cacheLock:            &sync.Mutex{},
+		cache:                make(map[core.RecordRef]cacheEntry, 0),
```
golangcibot: should use make(map[core.RecordRef]cacheEntry) instead (from `megacheck`)

ledger/artifactmanager/artifactmanager.go
```diff
@@ -97,6 +109,20 @@ func (m *LedgerArtifactManager) GetCode(
 	var err error
 	defer instrument(ctx, "GetCode").err(&err).end()
 
+	m.cacheLock.Lock()
+	entry, ok := m.cache[code]
+	if !ok {
+		m.cache[code] = cacheEntry{}
```
ruz: entry variable stays nil, you don't set it, so cache doesn't work.

ledger/artifactmanager/artifactmanager.go
```diff
@@ -124,6 +150,7 @@ func (m *LedgerArtifactManager) GetCode(
 		machineType: react.MachineType,
 	}
 	desc.cache.code = react.Code
```
ruz: Do we still need desc.cache.code? I don't think so.

ledger/artifactmanager/artifactmanager.go
```diff
@@ -40,10 +40,17 @@ type LedgerArtifactManager struct {
 	db                         *storage.DB
 	DefaultBus                 core.MessageBus                 `inject:""`
 	PlatformCryptographyScheme core.PlatformCryptographyScheme `inject:""`
+	cacheLock                  *sync.Mutex
+	cache                      map[core.RecordRef]cacheEntry
```
ruz: FIeld name is too generic, after all it's only cache of code.

ledger/artifactmanager/artifactmanager.go
```diff
@@ -40,10 +40,17 @@ type LedgerArtifactManager struct {
...
 	DefaultBus                 core.MessageBus                 `inject:""`
 	PlatformCryptographyScheme core.PlatformCryptographyScheme `inject:""`
+	codeCacheLock              *sync.Mutex
+	codeCache                  map[core.RecordRef]*cacheEntry
 
 	getChildrenChunkSize int
 }
 
+type cacheEntry struct {
+	sync.Mutex
```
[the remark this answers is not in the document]
afiskon: After a brief description we agreed that this approach is OK since this lock prevents sending multiple requests. However a corresponding comment should be added here.

PR #813, opened 2018-12-17
Ins 983 benchmark add api flag

megge-dream:
**- What I did**
add api url flag for benchmark


cmd/benchmark/main.go
```diff
@@ -157,6 +160,9 @@ func startScenario(s scenario) {
 func main() {
 	parseInputParams()
 
+	err := log.SetLevel("error")
```
ivanshibitov: add additional flag for changing of log level

PR #814, opened 2018-12-17
INS-25. Split.

andreyromancev:
Closes INS-25.
**- What I did**
Implemented jet splitting.
**- How to verify it**
Run tests.


ledger/pulsemanager/pulsemanager.go
```diff
@@ -258,28 +268,82 @@ func (m *PulseManager) processRecentObjects(
...
+	return msg, nil
+}
+
+func (m *PulseManager) sendExecutorData(
+	ctx context.Context,
+	currentPulse, newPulse *core.Pulse,
+	jetID core.RecordID,
+	msg *message.HotData,
+) error {
+	shouldSplit := func() (bool, error) {
```
nordicdyno: 1. error is not possible here, why function still returns it?
2. I suggest to move split logic in separate function with nice godoc and test

PR #853, opened 2018-12-19
INS-1008 resolve deadlock when we are new executor

aalekseev-ru:
**- What I did**
resolve deadlock
**- How I did it**
don't send message on next executor if it is me
**- How to verify it**
run tests

### DIFF
--- a/logicrunner/logicrunner.go
+++ b/logicrunner/logicrunner.go
@@ -291,8 +291,8 @@ func (lr *LogicRunner) Stop(ctx context.Context) error {
 func (lr *LogicRunner) CheckOurRole(ctx context.Context, msg core.Message, role core.DynamicRole) error {
 	// TODO do map of supported objects for pulse, go to jetCoordinator only if map is empty for ref
 	target := msg.DefaultTarget()
-	isAuthorized, err := lr.JetCoordinator.IsAuthorized(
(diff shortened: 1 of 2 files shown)


logicrunner/logicrunner.go
```diff
@@ -813,6 +813,14 @@ func (lr *LogicRunner) OnPulse(ctx context.Context, pulse core.Pulse) error {
...
 	// send copy for validation
 	for ref, state := range lr.state {
+		// we are executor again - we still working
+		// TODO we need to do something with validation
+		isAuthorized, _ := lr.JetCoordinator.AmI(
+			ctx, core.DynamicRoleVirtualExecutor, ref.Record(), pulse.PulseNumber)
+
+		if isAuthorized {
+			continue
+		}
```
afiskon: Maybe it should be `break` ?

logicrunner/logicrunner_unit_test.go
```diff
@@ -38,7 +44,7 @@ func TestOnPulse(t *testing.T) {
 	lr.state[objectRef] = &ObjectState{ExecutionState: &ExecutionState{Behaviour: &ValidationSaver{}}}
 	err = lr.OnPulse(ctx, pulse)
 	require.NoError(t, err)
-	require.Nil(t, lr.state[objectRef].ExecutionState)
+	assert.Nil(t, lr.state[objectRef].ExecutionState)
```
afiskon: require vs assert - what's the difference?

PR #856, opened 2018-12-20
INS-1022: don't apply loop detection to NoWait calls

ruz:
these were false positives: just call 1000 NoWaits on one object, they
all will be with the same TraceID and on executor and in one queue

### DIFF
--- a/logicrunner/logicrunner.go
+++ b/logicrunner/logicrunner.go
@@ -352,12 +352,8 @@ func (lr *LogicRunner) Execute(ctx context.Context, parcel core.Parcel) (core.Re
 		return nil, errors.Wrap(err, "can't play role")
 	}
 
-	if es.Current != nil {
-		// TODO: check no wait call
-		if inslogger.TraceID(es.Current.Context) == inslogger.TraceID(ctx) {
-			es.Unlock()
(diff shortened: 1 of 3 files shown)


logicrunner/logicrunner.go
```diff
@@ -352,12 +352,8 @@ func (lr *LogicRunner) Execute(ctx context.Context, parcel core.Parcel) (core.Re
 		return nil, errors.Wrap(err, "can't play role")
 	}
 
-	if es.Current != nil {
-		// TODO: check no wait call
-		if inslogger.TraceID(es.Current.Context) == inslogger.TraceID(ctx) {
-			es.Unlock()
-			return nil, os.WrapError(nil, "loop detected")
-		}
+	if loop := lr.CheckExecutionLoop(ctx, es, parcel); loop {
```
dmitry-panchenko: Why do you assign result of `CheckExecutionLoop ` to `loop` and never use it?

ruz: shit :) I use it right after ";" on the same line ^))))))))))

ruz: "fixed"

PR #879, opened 2018-12-20
NET-89: consensus packets signs

krestkrest:
**- What I did**
Fixed signing and signature verification in consensus packets
**- How to verify it**
run unit tests


consensus/packets/phase1packet.go
```diff
@@ -39,27 +40,63 @@ type Phase1Packet struct {
...
 	}
-	p1p.packetHeader.setRoutingFields(header, Phase1)
+	valid := crypto.Verify(key, core.SignatureFromBytes(p1p.Signature[:]), raw)
+	if !valid {
+		return errors.New("bad signature")
+	}
+	return nil
+}
 
+func (p1p *Phase1Packet) Sign(crypto core.CryptographyService) error {
```
dmitry-panchenko: Use `cryptographyService` name for `core.CryptographyService ` otherwise it's highly unclear and shadowing `crypto` package.

PR #904, opened 2018-12-25
 5 node net functors

ivanshibitov:
**- What I did**
5 node func tests

**- How I did it**

**- How to verify it**
Run tests

### DIFF
--- a/component/manager.go
+++ b/component/manager.go
@@ -92,6 +92,7 @@ func (m *Manager) Start(ctx context.Context) error {
 			if err != nil {
 				return errors.Wrap(err, "Failed to start components.")
 			}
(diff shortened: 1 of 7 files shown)


component/manager.go
```diff
@@ -92,6 +92,7 @@ func (m *Manager) Start(ctx context.Context) error {
 			if err != nil {
 				return errors.Wrap(err, "Failed to start components.")
 			}
+			log.Debugf("ComponentManager: Component %s started ", name)
 		} else {
 			log.Debug("ComponentManager: Component %s has no Start method", name)
```
LinnikD: use Debugf here pls

PR #917, opened 2018-12-26
NOISSUE: err->string

megge-dream:
**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/logicrunner/logicrunner.go
+++ b/logicrunner/logicrunner.go
@@ -105,10 +105,6 @@ type Error struct {
 	Method   string
 }
 
-func init() {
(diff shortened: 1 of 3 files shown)


logicrunner/logicrunner.go
```diff
@@ -105,6 +105,10 @@ type Error struct {
 	Method   string
 }
 
+func init() {
+	gob.Register(&Error{})
+}
+
```
afiskon: This code looks redundant.

megge-dream: Removed it

PR #971, opened 2019-01-10
INS-1111: api timeout

ilyapt:
**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/api/call.go
+++ b/api/call.go
@@ -36,10 +36,6 @@ import (
 	"github.com/pkg/errors"
 )
 
-const (
-	callTimeout int = 50 // seconds
-)
-
 var scheme = platformpolicy.NewPlatformCryptographyScheme()
 
 // Request is a representation of request struct to api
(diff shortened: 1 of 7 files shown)


api/call.go
```diff
@@ -168,15 +186,11 @@ func (ar *Runner) callHandler() func(http.ResponseWriter, *http.Request) {
 
 				insLog.Info("[ callHandler ] Incoming request: %s", req.RequestURI)
 
+				var sended uint32 = 0
```
golangcibot: should drop = 0 from declaration of var sended; it is the zero value (from `golint`)

api/call.go
```diff
@@ -34,6 +36,10 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	callTimeout int = 50 // seconds
```
LinnikD: move to config, default 2 seconds

PR #988, opened 2019-01-12
Noissue light check role in genesis

ivanshibitov:
**- What I did**
Add light check of existence of 3 different roles in genesis config

**- How to verify it**
Run tests

### DIFF
--- a/genesis/config.go
+++ b/genesis/config.go
@@ -39,7 +39,7 @@ func hasMinimumRolesSet(conf *genesisConfig) error {
 
 	if len(minRequiredRolesSet) != 0 {
 		var missingRoles string
-		for role, _ := range minRequiredRolesSet {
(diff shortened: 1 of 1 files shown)


genesis/config.go
```diff
@@ -25,6 +25,29 @@ type genesisConfig struct {
...
+		"light_material": true,
+	}
+
+	for _, discNode := range conf.DiscoveryNodes {
+		delete(minRequiredRolesSet, discNode.Role)
+	}
+
+	if len(minRequiredRolesSet) != 0 {
+		var missingRoles string
+		for role, _ := range minRequiredRolesSet {
```
golangcibot: should omit 2nd value from range; this loop is equivalent to `for role := range ...` (from `golint`)

PR #1009, opened 2019-01-16
NOISSUE: added exporter error code

ita-sammann:
**- What I did**
Added special error code in exporter.Export when pulse is not found

### DIFF
--- a/api/exporter.go
+++ b/api/exporter.go
@@ -19,7 +19,9 @@ package api
 import (
 	"context"
 	"net/http"
+	"strings"
 
+	jsonrpc "github.com/gorilla/rpc/v2/json2"
 	"github.com/insolar/insolar/core"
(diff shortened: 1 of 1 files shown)


api/exporter.go
```diff
@@ -84,6 +86,13 @@ func (s *StorageExporterService) Export(r *http.Request, args *StorageExporterAr
 	ctx := context.TODO()
 	result, err := exp.Export(ctx, core.PulseNumber(args.From), args.Size)
 	if err != nil {
+		if strings.Contains(err.Error(), "failed to fetch pulse data") {
+			return &jsonrpc.Error{
+				Code:    10404,
```
dmitry-panchenko: Shouldn't we extract error code to constant?

ita-sammann: No

PR #1021, opened 2019-01-17
9 nodes with 5 node tests

LinnikD:
**- What I did**

9 nodes locally


functest/main_test.go
```diff
@@ -125,31 +123,11 @@ func loadRootKeys() error {
 }
 
 func setInfo() error {
-	jsonValue, err := json.Marshal(postParams{
-		"jsonrpc": "2.0",
-		"method":  "info.Get",
-		"id":      "",
-	})
+	response, err := requester.Info(TestAPIURL)
```
megge-dream: Why not use info?

PR #1024, opened 2019-01-17
es.Current is a pointer, es is a pointer, protect from change under us

ruz:
it shouldn't happen, but it feels that some code replaces es.Current
while we are executing "call constructor". We start with one es.Current.Request
and return different one. Protect this code by making a copy of
*es.Current into a local variable, so data stays the same within one
function. Do it for call constructor and for call method as well.

### DIFF
--- a/logicrunner/logicrunner.go
+++ b/logicrunner/logicrunner.go
@@ -72,7 +72,6 @@ type ExecutionState struct {
 }
 
 type CurrentExecution struct {
-	sync.Mutex
 	Context       context.Context
 	LogicContext  *core.LogicCallContext
 	Request       *Ref
--- a/logicrunner/logicrunner.go
+++ b/logicrunner/logicrunner.go
@@ -676,13 +676,17 @@ func (lr *LogicRunner) prepareObjectState(ctx context.Context, msg *message.Exec
 
 	es.Lock()
 
-	// prepare pending
-	if es.pending == message.PendingUnknown {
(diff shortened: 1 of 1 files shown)


logicrunner/logicrunner.go
```diff
@@ -737,9 +737,10 @@ func (lr *LogicRunner) executeMethodCall(ctx context.Context, es *ExecutionState
 		inslogger.FromContext(ctx).Info("LogicRunner.executeMethodCall starts")
 	}
 
-	es.Current.LogicContext.Prototype = es.objectbody.Prototype
-	es.Current.LogicContext.Code = es.objectbody.CodeRef
-	es.Current.LogicContext.Parent = es.objectbody.Parent
+	current := *es.Current
```
golangcibot: assignment copies lock value to current: github.com/insolar/insolar/logicrunner.CurrentExecution (from `govet`)

logicrunner/logicrunner.go
```diff
@@ -836,23 +837,24 @@ func (lr *LogicRunner) executeConstructorCall(
 ) (
 	core.Reply, error,
 ) {
-	if es.Current.LogicContext.Caller.IsEmpty() {
+	current := *es.Current
```
golangcibot: assignment copies lock value to current: github.com/insolar/insolar/logicrunner.CurrentExecution (from `govet`)

PR #1034, opened 2019-01-18
INS-1140: benchmark bank test

afiskon:


api/sdk/sdk.go
```diff
@@ -182,3 +182,29 @@ func (sdk *SDK) Transfer(amount uint, from *Member, to *Member) (string, error)
 
 	return response.TraceID, nil
 }
+
+func (sdk *SDK) GetBalance(m *Member) (uint64, error) {
```
megge-dream: Add comment to this func

megge-dream: Why uint64, and not uint?

afiskon: Done.

afiskon: In fact I would prefer to use some sort of arbitrary precision numbers since we never told that balance is limited to 2^64 coins or any other number. But since currently we don't use arbitrary precision numbers for anything I chose an unsigned integer with largest precision.

PR #1057, opened 2019-01-18
any number of nodes in launchnet

ivanshibitov:
**- What I did**

**- How I did it**

**- How to verify it**


genesis/config.go
```diff
@@ -5,14 +5,16 @@ import (
...
+type Discovery struct {
 	Host     string `mapstructure:"host"`
 	Role     string `mapstructure:"role"`
 	KeysFile string `mapstructure:"keys_file"`
 	CertName string `mapstructure:"cert_name"`
 }
 
-type genesisConfig struct {
+// GenesisConfig contains all genesis config
+type GenesisConfig struct {
```
golangcibot: type name will be used as genesis.GenesisConfig by other packages, and that stutters; consider calling this Config (from `golint`)

PR #1058, opened 2019-01-18
it's possible that a message stays in message bus queue for a pulse

ruz:
checkPulse thinks that it's a message from future, but pulse changes
before it can block on old pulse channel (barrier), we already swiped it
with a new one, so message stays in the queue for pulse duration

### DIFF
--- a/messagebus/messagebus.go
+++ b/messagebus/messagebus.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"encoding/gob"
 	"fmt"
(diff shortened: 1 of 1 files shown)


messagebus/messagebus.go
```diff
@@ -21,6 +21,7 @@ import (
 	"context"
 	"encoding/gob"
 	"fmt"
+	"github.com/insolar/insolar/core/utils"
```
aMFeriuSS: This import is not in its place

PR #1111, opened 2019-01-21
NOISSUE: fix pulsar distribution

krestkrest:
**- What I did**
1. Fixed bug in pulsar distributor that caused timeouts
2. Now pulsar sends pulses only to the nodes from `pulsedistributor.bootstraphosts` configuration params (getRandomHosts is not called)
3. Added magic number `Skip` to servicenetwork configuration that allows to skip a couple of pulses before network bootstrap to have synchronized first pulse on all nodes (very helpful on short pulses < 3000ms)
4. Pulsar config in launchnet is now generated based on `pulsar_template.yaml` (all settings except bootstraphosts) and `genesis.yaml` (bootstraphosts)

**- How to verify it**
Run launchnet.sh on 20+ nodes and check with script `ag -m1 'persist' scripts/insolard/nodes/*/output.log 2>/dev/null | awk '{print $4}' | sort -u` that all nodes start from one pulse

### DIFF
--- a/cmd/insolard/main.go
+++ b/cmd/insolard/main.go
@@ -112,7 +112,7 @@ func main() {
 		bootstrapComponents.KeyProcessor,
 	)
 
-	fmt.Print("Starts with configuration:\n", configuration.ToString(cfgHolder.Configuration))
+	fmt.Println("Starts with configuration:\n", configuration.ToString(cfgHolder.Configuration))
 
 	jaegerflush := func() {}
 	if params.traceEnabled {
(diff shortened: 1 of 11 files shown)


metrics/network.go
```diff
@@ -67,3 +67,11 @@ var NetworkConnections = prometheus.NewGauge(prometheus.GaugeOpts{
 	Namespace: insolarNamespace,
 	Subsystem: "network",
 })
+
+// NetworkConnections is current network transport connections count metric
```
golangcibot: comment on exported var NetworkComplete should be of the form "NetworkComplete ..." (from `golint`)

PR #1133, opened 2019-01-22
NOISSUE: send metric to prometheus when node reaches complete network state

krestkrest:


metrics/network.go
```diff
@@ -67,3 +67,11 @@ var NetworkConnections = prometheus.NewGauge(prometheus.GaugeOpts{
 	Namespace: insolarNamespace,
 	Subsystem: "network",
 })
+
+// NetworkConnections is current network transport connections count metric
```
golangcibot: comment on exported var NetworkComplete should be of the form "NetworkComplete ..." (from `golint`)

PR #1138, opened 2019-01-23
Add traces and make addSync and clean async.

andreyromancev:


ledger/heavyclient/heavy_client.go
```diff
@@ -149,7 +149,7 @@ func (c *JetClient) runOnce(ctx context.Context) {
 	c.startOnce.Do(func() {
 		// TODO: reset TraceID from context, or just don't use context?
 		// (TraceID not meaningful in async sync loop)
-		ctx, cancel := context.WithCancel(ctx)
+		ctx, cancel := context.WithCancel(context.Background())
```
nordicdyno: we need to pass logger in this context

nordicdyno: nope, the global logger is fine here

PR #1157, opened 2019-01-24
Ins 1197 split db to interfaces

egorikas:
**- What I did**
There are a few new interfaces. So, there is no need anymore to use `db` in methods

**- How I did it**

**- How to verify it**


ledger/jetcoordinator/jetcoordinator.go
```diff
@@ -32,17 +32,18 @@ import (
 
 // JetCoordinator is responsible for all jet interactions
 type JetCoordinator struct {
-	db                         *storage.DB
+	roleCounts                 map[core.DynamicRole]int
```
golangcibot: `roleCounts` is unused (from `structcheck`)

PR #1162, opened 2019-01-28
INS-1167: add error for exporter api when fromPulse bigger then current

aMFeriuSS:
**- What I did**
Error for exporter api when fromPulse bigger then current was added
**- How I did it**
```
if fromPulsePN > currentPulse.PulseNumber {
	return nil, errors.New("failed to fetch data: from-pulse > current-pulse")
}
```


aMFeriuSS:
> could you add a test for this case?
Done.

```	
_, err = exporter.Export(ctx, 100000, 2)
require.Error(t, err, "From-pulse should be smaller (or equal) current-pulse")

_, err = exporter.Export(ctx, 60000, 2)
require.NoError(t, err, "From-pulse should be smaller (or equal) current-pulse")
```

PR #1171, opened 2019-01-29
Ins 1203 middleware am

egorikas:
**- What I did**

**- How I did it**

**- How to verify it**


ledger/artifactmanager/artifactmanager_senders.go
```diff
@@ -0,0 +1,156 @@
...
+) *ledgerArtifactSenders {
+	return &ledgerArtifactSenders{
+		pulseStorage: pulseStorage,
+		jetStorage:   jetStorage,
+		defaultBus:   defaultBus,
+		codeCache:    map[core.RecordRef]*cacheEntry{},
+	}
+}
+
+func (m *ledgerArtifactSenders) bus(ctx context.Context) core.MessageBus {
```
andreyromancev: This should probably be passed as an argument to PreSender constructor. This way, the PreSender won't have to know how to get the bus. We construct PreSenders every time before send, so we can just store MessageBus reference in it and get rid of this method.

ledger/artifactmanager/artifactmanager_senders.go
```diff
@@ -0,0 +1,156 @@
...
+		if !ok {
+			return response, err
+		}
+
+		entry.code = castedResp
+		return response, err
+	}
+}
+
+func (m *ledgerArtifactSenders) followRedirectSender(sender Sender) Sender {
```
andreyromancev: If can be a function with a closure on MessageBus, passed as an argument (see instrumentHandler).

ledger/artifactmanager/artifactmanager_senders.go
```diff
@@ -0,0 +1,156 @@
...
+				return nil, errors.New("double redirects are forbidden")
+			}
+			return rep, nil
+		}
+
+		return rep, err
+	}
+}
+
+func (m *ledgerArtifactSenders) retryJetSender(sender Sender) Sender {
```
andreyromancev: If can be a function with a closure on JetStorage and PulseNumber, passed as an argument (see instrumentHandler).

ledger/artifactmanager/artifactmanager_senders.go
egorikas: After conversation there is a decision not to do this action.

ledger/artifactmanager/artifactmanager_senders.go
```diff
@@ -30,127 +30,123 @@ import (
...
 
-		currentPulse, err := m.pulseStorage.Current(ctx)
-		if err != nil {
-			return nil, err
-		}
+			retries := jetMissRetryCount
 
-		for retries > 0 {
-			rep, err := sender(ctx, msg, options)
+			currentPulse, err := pulseStorage.Current(ctx)
```
andreyromancev: We need to pass PulseNumber here, not PulseStorage. When conveyer comes, all calls to PulseStorage would be forbidden.

ledger/artifactmanager/artifactmanager_senders.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/insolar/insolar/instrumentation/inslogger"
+	"github.com/insolar/insolar/ledger/storage"
+	"github.com/pkg/errors"
+	"go.opencensus.io/stats"
+)
+
+// ledgerArtifactSenders is a some kind of a middleware layer
+// it contains cache meta-data for calls
+type ledgerArtifactSenders struct {
+	codeCacheLock sync.Mutex
```
andreyromancev: TO make this universal, we need to use generic cache, not codeCache and use message hash as key. If its too hard or not required for now, let's leave a TODO comment.

PR #1179, opened 2019-01-30
INS-1235: do not send state back to light, if it wasn't changed

bigbes:
* also added mocks for MachineLogicExecutor

**- What I did**
if state wasn't changed (cbor byte representation is the same), then we won't create new amend for current object.

**- How I did it**

**- How to verify it**
test included

### DIFF
--- a/logicrunner/logicrunner.go
+++ b/logicrunner/logicrunner.go
@@ -798,7 +798,7 @@ func (lr *LogicRunner) executeMethodCall(ctx context.Context, es *ExecutionState
 		if err != nil {
 			return nil, es.WrapError(err, "couldn't deactivate object")
 		}
-	} else {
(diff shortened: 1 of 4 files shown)


logicrunner/logicrunner.go
```diff
@@ -798,7 +798,7 @@ func (lr *LogicRunner) executeMethodCall(ctx context.Context, es *ExecutionState
 		if err != nil {
 			return nil, es.WrapError(err, "couldn't deactivate object")
 		}
-	} else {
+	} else if bytes.Compare(es.objectbody.Object, newData) != 0 {
```
golangcibot: should use !bytes.Equal(es.objectbody.Object, newData) instead (from `gosimple`)

PR #1191, opened 2019-02-01
INS-1307: discovery restart

krestkrest:
**- What I did**
Discovery nodes can now return to running network after shutdown
**- How to verify it**
1. Run integration test TestDiscoveryRestart
2. Run launchnet.sh, shut down one of the discovery nodes and then start it again


network/interfaces.go
```diff
@@ -191,6 +192,8 @@ type NodeKeeper interface {
 	ResolveConsensus(shortID core.ShortNodeID) *host.Host
 	// ResolveConsensusRef get temporary mapping by node ID
 	ResolveConsensusRef(nodeID core.RecordRef) *host.Host
+	// Wipe all active nodes for test purposes
+	Wipe(isDiscovery bool)
```
AndreyBronin: remove it from the interface, use cast to impl in test

PR #1197, opened 2019-02-01
INS-1332: node start script

ooki2day:


scripts/generate_insolar_configs.go
```diff
@@ -223,7 +225,8 @@ func main() {
 		pulsarConfig.Pulsar.PulseDistributor.BootstrapHosts = append(pulsarConfig.Pulsar.PulseDistributor.BootstrapHosts, node.Host)
 	}
 
-	writeInsolarConfigs(insolarConfigs)
+	writeInsolarConfigs(outputDir, discoveryNodesConfigs)
+	writeInsolarConfigs(outputDir+"/nodes", nodesConfigs)
```
krestkrest: we will have discovery nodes configs in directory `discovery_nodes` and configs for joiner nodes in `discovery_nodes/nodes`, is it desired behavior?

AndreyBronin: use https://golang.org/pkg/path/filepath/ join to work with paths

ooki2day: fixed

PR #1207, opened 2019-02-05
INS-1348: jet trees in memory

ruz:
no point for lights materials to store anything on disk, use in memory
trees.


ledger/storage/db.go
```diff
@@ -114,6 +114,7 @@ type DB struct {
 	idlocker             *IDLocker
 	jetHeavyClientLocker *IDLocker
 
+	jetTree          map[core.PulseNumber]*jet.Tree
```
egorikas: Why do we need it here?

egorikas: Why do we need it here?

andreyromancev: It should go to `jetStorage` struct

ruz: done

PR #1208, opened 2019-02-05
INS-1369 returnResults timeout

aalekseev-ru:
**- What I did**
clean CR channels map if returnResults doesn't come in time
**- How I did it**
add context.WithTimeout, refactor code for taking timeout from config
**- How to verify it**
run tests

### DIFF
--- a/configuration/apirunner.go
+++ b/configuration/apirunner.go
@@ -18,6 +18,7 @@ package configuration
 
 import (
 	"fmt"
+	"time"
 )
 
 // APIRunner holds configuration for api
@@ -25,7 +26,7 @@ type APIRunner struct {
 	Address string
 	Call    string
 	RPC     string
-	Timeout uint32
+	Timeout time.Duration
 }
 
 // NewAPIRunner creates new api config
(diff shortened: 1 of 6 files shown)


configuration/apirunner.go
```diff
@@ -18,14 +18,15 @@ package configuration
...
+	"time"
 )
 
 // APIRunner holds configuration for api
 type APIRunner struct {
 	Address string
 	Call    string
 	RPC     string
-	Timeout uint32
+	Timeout time.Duration
```
Vany: dimension of this value was seconds, now it microseconds. It will be misleading in configs.

contractrequester/contractrequester.go
```diff
@@ -145,6 +146,8 @@ func (cr *ContractRequester) CallMethod(ctx context.Context, base core.Message,
 		return res, nil
 	}
 
+	ctx, cancel := context.WithTimeout(ctx, configuration.NewAPIRunner().Timeout)
```
Vany: we have same situation in callconstructor, add it there too, pls.

configuration/apirunner.go
aalekseev-ru: fix

contractrequester/contractrequester.go
aalekseev-ru: added, thinling about test suites

PR #1213, opened 2019-02-06
INS-1405: Don’t rewrite pending requests.

andreyromancev:


ledger/pulsemanager/pulsemanager.go
```diff
@@ -168,11 +161,23 @@ func (m *PulseManager) processEndPulse(
 			}
 
 			if info.left == nil && info.right == nil {
+				msg, err := m.getExecutorHotData(
```
nordicdyno: why not move duplicate code outside of `if` block

PR #1217, opened 2019-02-06
INS-1263: Consensus metrics

ooki2day:


consensus/phases/communicator.go
```diff
@@ -371,6 +376,7 @@ func (nc *NaiveCommunicator) sendAdditionalRequests(origReq *packets.Phase2Packe
 		newReq := *origReq
 		newReq.AddVote(&packets.MissingNode{NodeIndex: uint16(req.RequestIndex)})
 		receiver := selectCandidate(req.Candidates)
+		metrics.ConsensusPacketsSent.WithLabelValues("Additional request", "to: "+receiver.String())
```
krestkrest: No finalizing expression, e.g. `Inc`

consensus/phases/communicator.go
```diff
@@ -257,6 +260,7 @@ func (nc *NaiveCommunicator) ExchangePhase1(
 	for {
 		select {
 		case res := <-nc.phase1result:
+			metrics.ConsensusPacketsRecv.WithLabelValues("phase 1").Inc()
```
krestkrest: I don't think that we should count consensus packets here because they can be filtered later. It is better to record all the packets after `ExchangePhaseN` in the phase code (where we have log record `received packets: %d/%d`)

consensus/phases/communicator.go
```diff
@@ -417,6 +423,7 @@ func (nc *NaiveCommunicator) ExchangePhase21(ctx context.Context, list network.U
 	for {
 		select {
 		case res := <-nc.phase2result:
+			metrics.ConsensusPacketsRecv.WithLabelValues("phase 2").Inc()
```
krestkrest: see https://github.com/insolar/insolar/pull/1217#discussion_r254375165

consensus/phases/communicator.go
```diff
@@ -483,6 +490,7 @@ func (nc *NaiveCommunicator) ExchangePhase3(ctx context.Context, participants []
 	for {
 		select {
 		case res := <-nc.phase3result:
+			metrics.ConsensusPacketsRecv.WithLabelValues("phase 3").Inc()
```
krestkrest: see https://github.com/insolar/insolar/pull/1217#discussion_r254375165

consensus/phases/communicator.go
```diff
@@ -324,6 +328,7 @@ func (nc *NaiveCommunicator) ExchangePhase2(ctx context.Context, list network.Un
 	for {
 		select {
 		case res := <-nc.phase2result:
+			metrics.ConsensusPacketsRecv.WithLabelValues("phase 2").Inc()
```
krestkrest: see https://github.com/insolar/insolar/pull/1217#discussion_r254375165

consensus/phases/communicator.go
ooki2day: done

consensus/phases/communicator.go
ooki2day: done

consensus/phases/communicator.go
ooki2day: done

consensus/phases/communicator.go
ooki2day: done

consensus/phases/communicator.go
ooki2day: done

PR #1236, opened 2019-02-08
Ins 1378 nodes in genesis

megge-dream:
**- What I did**
You can reuse keys for testing

**- How I did it**
Add prepare logic in genesis

**- How to verify it**
Run tests


genesis/helpers.go
```diff
@@ -104,3 +104,11 @@ func getKeysFromFile(ctx context.Context, file string) (crypto.PrivateKey, strin
 	}
 	return key, keys["public_key"], nil
 }
+
+func abs(path string) (string, error) {
```
ivanshibitov: this name looks like this function return absolute value of number

genesis/genesis.go
```diff
@@ -379,6 +376,177 @@ func (g *Genesis) activateDiscoveryNodes(ctx context.Context, cb *ContractsBuild
...
+
+		result, err := json.MarshalIndent(map[string]interface{}{
+			"private_key": string(privKeyStr),
+			"public_key":  string(pubKeyStr),
+		}, "", "    ")
+		if err != nil {
+			return errors.Wrap(err, "[ createKeys ] couldn't marshal keys")
+		}
+
+		name := "node_" + strconv.Itoa(i) + ".json"
```
ivanshibitov: Is this format used anywhere?
I mean "node-<id>.json".
I thing yes, that is why it would be better use format string for that

genesis/genesis.go
```diff
@@ -379,6 +376,177 @@ func (g *Genesis) activateDiscoveryNodes(ctx context.Context, cb *ContractsBuild
...
+		}
+	}
+
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		return nil, errors.Wrap(err, "[ uploadKeys ] dir is not exist")
+	}
+
+	files, _ := ioutil.ReadDir(path)
+	if len(files) != amount {
+		return nil, errors.New("[ uploadKeys ] amount of nodes != amount of files in directory: " + strconv.Itoa(len(files)) + "!=" + strconv.Itoa(amount))
```
ivanshibitov: here It's better to use fmt.Sprintf() , it's more understandable

PR #1249, opened 2019-02-11
INS-1439: Generate record types.

andreyromancev:


ledger/storage/record/registry.go
```diff
@@ -0,0 +1,21 @@
...
+func register(id TypeID, r Record) {
+	if _, ok := registry[id]; ok {
+		panic("duplicate record type")
+	}
+
+	registry[id] = r
+}
+
+// Registry returns records by type.
+func Registry() map[TypeID]Record {
```
egorikas: I suppose, it's better for us to rename this method. Smth like GetRegistrated or Registrated

ledger/storage/record/record.go
```diff
@@ -21,15 +21,30 @@ import (
...
+
+	register(20, new(RequestRecord))
+
+	register(30, new(ResultRecord))
+	register(31, new(TypeRecord))
+	register(32, new(CodeRecord))
+	register(33, new(ObjectActivateRecord))
+	register(34, new(ObjectAmendRecord))
+	register(35, new(DeactivationRecord))
+}
```
ita-sammann: I suggest to use hundreds or even thousands here.

ita-sammann: And also you should note here, that these are not magic numbers. This numbers haven't any meaning

PR #1269, opened 2019-02-13
INS-1489: Remove sync from config.

andreyromancev:


andreyromancev:
Floating test, merging.

PR #1272, opened 2019-02-13
INS-1459: discovery restart testcase

krestkrest:
**- What I did**
Fixed testcase when discovery node stops and tries to reconnect to the network. Also passed context from consensus to consensus communicator and added more logs. 
**- How to verify it**
Run integration tests and launchnet.sh
**- QA notes**
There is still a bug that the stopped node should be restarted not earlier than two pulses from its stop (else the node will get stuck). This bug will be fixed in a separate task.


network/servicenetwork/integr_test.go
```diff
@@ -1,5 +1,3 @@
-// +build networktest
```
dmitry-panchenko: Don't do that

PR #1274, opened 2019-02-13
INS-1435: Search proxy and go files in multiple directories

bigbes:
Sometimes tests were runned from root folder, but sometimes
they were runned form logicrunner folder. That (relative path
for test fixture) was the reason for test to silently crash in
non-related place.

Now tests are searching in root folder and in LogicRunner folder

### DIFF
--- a/logicrunner/logicrunner_test.go
+++ b/logicrunner/logicrunner_test.go
@@ -24,6 +24,7 @@ import (
 	"io/ioutil"
 	"net/rpc"
 	"os"
+	"path"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


logicrunner/logicrunner_test.go
```diff
@@ -1487,24 +1502,37 @@ func (r *One) CreateAllowance(member string) (error) {
...
+		memberCodePath = path.Join("..", memberCodePath)
+		allowanceCodePath = path.Join("..", allowanceCodePath)
+		walletCodePath = path.Join("..", walletCodePath)
+
+		if _, err := os.Stat(rootDomainPath); os.IsNotExist(err) {
+			t.Fatal("Failed to find rootDomain file in ", rootDomainPath)
+		}
+	}
+
+	rootDomainCode, err := ioutil.ReadFile(rootDomainPath)
```
ita-sammann: Maybe init all these paths before all tests to avoid duplication?

PR #1292, opened 2019-02-15
INS-1495+INS-1435, LogicRunner: tests to suites + contracts now searched in right place in tests

bigbes:


bigbes:
Obsoletes https://github.com/insolar/insolar/pull/1274

logicrunner/logicrunner_test.go
```diff
@@ -63,25 +68,44 @@ import (
...
+
+func FindContractsDir() (string, error) {
+	_, file, _, ok := runtime.Caller(0)
+	if !ok {
+		return "", errors.New("failed to find folder")
+	}
+
+	// we're located in /logicrunner/logicrunner_test.go file, so we must take two basenames
+	projectRoot := path.Dir(path.Dir(file))
+	contractsDir := path.Join(projectRoot, "application", "contract")
```
aalekseev-ru: why not GetRealApplicationDir ?

logicrunner/logicrunner_test.go
```diff
@@ -233,9 +257,9 @@ func ValidateAllResults(t testing.TB, ctx context.Context, lr core.LogicRunner,
 
 		_, err := rlr.Validate(ctx, ref, *rlr.pulse(ctx), *cb)
 		if _, ok := failmap[ref]; ok {
-			assert.Error(t, err, "validation %s", ref)
+			t.Error(err, "validation %s", ref)
 		} else {
-			assert.NoError(t, err, "validation %s", ref)
+			t.Error(err, "validation %s", ref)
```
aalekseev-ru: NoError becomes Error?

logicrunner/logicrunner_test.go
bigbes: `GetRealApplicationDir` scans every directory in `GOPATH` and tries to deduce what is the insolar dir.
Better way is to read current directory of current file and that'll be project path (and that we won't look into another copy of insolar repo)
But I can't modify `GetRealApplicationDir`, since it's used in code, not only in tests.

logicrunner/logicrunner_test.go
bigbes: My mistake

PR #1312, opened 2019-02-19
INS-1238: refactoring AM with new method - Pulse(...)

aMFeriuSS:
**- What I did**
New method named Pulse added for ArtifactManager.
m.PulseManager.Current() replaced with m.Pulse()
**- How I did it**
```
func (m *LedgerArtifactManager) Pulse(ctx context.Context) (pn core.PulseNumber, err error) {
	pulse, err := m.PulseStorage.Current(ctx)
	if err != nil {
		return
	}

	pn = pulse.PulseNumber
	return
}
```

### DIFF
--- a/ledger/artifactmanager/middleware.go
+++ b/ledger/artifactmanager/middleware.go
@@ -58,9 +58,9 @@ func newMiddleware(
 
 func (m *middleware) addFieldsToLogger(handler core.MessageHandler) core.MessageHandler {
 	return func(ctx context.Context, parcel core.Parcel) (core.Reply, error) {
-		cntext, _ := inslogger.WithField(ctx, "targetid", parcel.DefaultTarget().String())
+		ctx, _ = inslogger.WithField(ctx, "targetid", parcel.DefaultTarget().String())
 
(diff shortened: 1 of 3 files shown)


ledger/artifactmanager/middleware.go
```diff
@@ -58,9 +58,9 @@ func newMiddleware(
 
 func (m *middleware) addFieldsToLogger(handler core.MessageHandler) core.MessageHandler {
 	return func(ctx context.Context, parcel core.Parcel) (core.Reply, error) {
-		context, _ := inslogger.WithField(ctx, "targetid", parcel.DefaultTarget().String())
+		cntext, _ := inslogger.WithField(ctx, "targetid", parcel.DefaultTarget().String())
```
andreyromancev: Why?

aMFeriuSS: IDE inspection:
```
Variable 'context' collides with imported package name less... (⌘F1)
Inspection info: Reports declarations of variables, arguments or functions 
that overlap with the used import
```

PR #1329, opened 2019-02-20
INS-1555: LocalStorage removing

aMFeriuSS:
**- What I did**
Remove LocalStorage (and localstorage package)


messagebus/player.go
```diff
@@ -58,7 +57,7 @@ func (p *player) Send(ctx context.Context, msg core.Message, ops *core.MessageSe
 
 	item, err := p.tape.Get(ctx, id)
 	if err != nil {
-		if err == localstorage.ErrNotFound {
+		if err == core.ErrNotFound {
```
andreyromancev: This is never. It's not used in the memory tape. Let's remove this condition.

aMFeriuSS: done

PR #1335, opened 2019-02-20
Ins 1446 zerolog adapter

ruz:


log/zerolog.go
```diff
@@ -0,0 +1,139 @@
...
+	switch strings.ToLower(cfg.Formatter) {
+	case "text":
+		output = zerolog.ConsoleWriter{Out: os.Stderr, NoColor: true, TimeFormat: timestampFormat}
+	case "json":
+		output = os.Stderr
+	default:
+		return nil, errors.New("unknown formatter " + cfg.Formatter)
+	}
+
+	zerolog.CallerSkipFrameCount = 4
```
nordicdyno: Could you add some comment with description why 4 or link on doc where this description present?

log/log.go
```diff
@@ -28,6 +28,7 @@ import (
 )
 
 const defaultSkipCallNumber = 3
+const timestampFormat = "2006-01-02 15:04:05.000000"
```
nordicdyno: Why custom format? Why not  RFC 3339?

log/log.go
```diff
@@ -28,6 +28,7 @@ import (
 )
 
 const defaultSkipCallNumber = 3
```
nordicdyno: why 3?

ruz: added test to show that it is correct

log/log.go
ruz: we want sub seconds

log/zerolog.go
ruz: no doc, no description, added test to show that it is correct

PR #1367, opened 2019-02-22
INS-1624: Added fixed address resolver for external ip substitution

bigbes:


krestkrest:
Need to know: these changes will affect only the main transport and will not affect consensus transport

PR #1373, opened 2019-02-25
INS-1460: make create_member single command

Vany:
./bin/insolar -c create_member <username>

### DIFF
--- a/cmd/insolar/main.go
+++ b/cmd/insolar/main.go
@@ -38,7 +38,8 @@ import (
 )
 
 const defaultStdoutPath = "-"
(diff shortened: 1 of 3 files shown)


cmd/insolar/main.go
```diff
@@ -116,6 +117,90 @@ func parseInputParams() {
 
 }
 
+func main() {
+	if u := os.Getenv("INSOLAR_API_URL"); u != "" {
```
bigbes: What's the reason for creating env variable? It's simply configured in the cli using arguments

PR #1403, opened 2019-03-04
Ins 1845 updates for conveyor

megge-dream:
**- What I did**
- add InitiateShutdown method
- move AntiquePulseNumber const to core.Pulse
- get info about new future pulse number from futurePulseData
- if createElement returns error, push element back to EmptyList
- PreparePulse/ActivatePulse check only ShuttingDown states, which allows init conveyer by doing PreparePulse and ActivatePulse calls 
- stateMachine is field in Slot struct now
- if we can't push signal into slots queue, panic

**- How to verify it**
Run tests

### DIFF
--- a/conveyor/conveyor.go
+++ b/conveyor/conveyor.go
@@ -159,12 +159,12 @@ func (c *PulseConveyor) SinkPushAll(pulseNumber core.PulseNumber, data []interfa
 
 // PreparePulse is preparing conveyor for working with provided pulse
 func (c *PulseConveyor) PreparePulse(pulse core.Pulse, callback queue.SyncDone) error {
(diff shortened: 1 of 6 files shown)


conveyor/conveyor.go
```diff
@@ -106,13 +107,22 @@ func (c *PulseConveyor) IsOperational() bool {
 	return false
 }
 
+func (c *PulseConveyor) InitiateShutdown(force bool) {
+	c.lock.Lock()
+	c.state = ShuttingDown
+	c.lock.Unlock()
+	if force {
```
golangcibot: SA9003: empty branch (from `staticcheck`)

PR #1411, opened 2019-03-06
Ins 1711 blob storage in memory

aMFeriuSS:


ledger/storage/blob/blob.go
```diff
@@ -0,0 +1,42 @@
...
+	"context"
+
+	"github.com/insolar/insolar/core"
+)
+
+// Accessor provides info about Blob-values from storage
+//go:generate minimock -i github.com/insolar/insolar/ledger/storage/blob.Accessor -o ./ -s _mock.go
+type Accessor interface {
+	// Get returns Blob for provided id
+	Get(ctx context.Context, id core.RecordID) (Blob, error)
```
andreyromancev: Let's call it "ForID", because there will be another one for jet.

ledger/storage/blob/blobmem.go
```diff
@@ -0,0 +1,79 @@
+/*
```
andreyromancev: This file doesn't have to begin with "blob". Let's call it just "memory" or "mem".

ledger/storage/blob/blob_cmp_test.go
```diff
@@ -0,0 +1,101 @@
...
+			resBlob, err := blobStorage.Get(ctx, bl.id)
+			require.NoError(t, err)
+
+			assert.Equal(t, bl.b, resBlob)
+			assert.Equal(t, bl.b.Value, resBlob.Value)
+			assert.Equal(t, bl.b.JetID, resBlob.JetID)
+		}
+	})
+
+	t.Run("returns override error when saving with the same id", func(t *testing.T) {
```
andreyromancev: This test fails if executed separately.

ledger/storage/blob/blob.go
aMFeriuSS: done

ledger/storage/blob/blobmem.go
aMFeriuSS: done

ledger/storage/blob/blob_cmp_test.go
aMFeriuSS: done

PR #1418, opened 2019-03-07
INS-1723. DB docs and tests.

andreyromancev:


ledger/storage/db/badger_test.go
```diff
@@ -1,17 +1,97 @@
...
 	t.Parallel()
 
-	// tmpdir, err := ioutil.TempDir("", "bdb-test-")
-	// assert.NoError(t, err)
-	//
-	// bdb, err := badger.Open(badger.Options{
-	// 	Dir:      tmpdir,
-	// 	ValueDir: tmpdir,
-	// })
+	tmpdir, err := ioutil.TempDir("", "bdb-test-")
```
aMFeriuSS: This tmp-dir should be removed.
Here citation from doc for  `ioutil.TempDir`
```
It is the caller's responsibility
to remove the directory when no longer needed.
```
(edit here and below)

PR #1419, opened 2019-03-07
Ins 1727 kube node

nordicdyno:


build/kube-node/templates/ve-sts.tmpl.yaml
```diff
@@ -0,0 +1,73 @@
...
+  replicas: 1
+  serviceName: insolar-ve
+  template:
+    metadata:
+      labels:
+        app: insolar-ve
+    spec:
+      containers:
+      - name: insgorund
+        image: "insolar/insgorund:v0.8.4"
```
krestkrest: move v0.8.4 to a separate variable in insgorund and insolard

PR #1445, opened 2019-03-15
NOISSUE: write generated code directly into the file

corpix:
This fixes code generation on systems where temporary files root is separate mount point


corpix:
Sure, I could use TMPDIR and it will work. But this is not comfortable, not cross-platform and I don't understand why combination of `ioutil.TempFile` + `os.Rename` should be used here.

PR #1451, opened 2019-03-18
INS-1947: Snapshot and Accessor interfaces/implementations

krestkrest:
**- What I did**
Replaced internal NodeKeeper state with Snapshot and Accessor (read interface for Snapshot). Also added some tests for accessor.
**- How to verify it**


network/nodenetwork/snapshot.go
```diff
@@ -17,9 +17,69 @@
...
+		pulse: number,
+		// TODO: pass actual state
+		state:    core.NoNetworkState,
+		nodeList: sortNodes(nodes),
+	}
+}
+
+// sortNodes temporary method to create snapshot lists. Will be replaced by special function that will take in count
+// previous snapshot and approved claims.
+func sortNodes(nodes map[core.RecordRef]core.Node) [ListLength][]core.Node {
```
AndreyBronin: may be "assort" or "separate" is better word for method name
There is no sorting in classic meaning here

krestkrest: renamed to "splitNodes"

PR #1453, opened 2019-03-19
INS-1862: give ability to execute request with needed logging level

bigbes:
* Extract logging level and give ability to copy logger
* Add logLevel to parcel, apply it automatically
* modified insolar/benchmark to use this log level (using loglevelserver
  flag)
* external passed logger level key is stored inside Context now
* testing
* extract data re request from parcel to subfield (for future extraction)


insolar/message/parcel.go
```diff
@@ -31,15 +31,21 @@ type ParcelFactory interface {
...
 
-// Parcel is a message signed by senders private key.
-type Parcel struct {
-	Sender        insolar.Reference
-	Msg           insolar.Message
-	Signature     []byte
+// ServiceData is a structure with utility fields like log level and trace id.
+type ServiceData struct {
 	LogTraceID    string
+	LogLevel      insolar.LogLevel
```
ruz: LogLevel field should be NILable. If it was not adjusted then it shouldn't affect any other node. Behaviour should be discussed in details.

insolar/logger.go
```diff
@@ -59,4 +120,7 @@ type Logger interface {
 	WithFields(map[string]interface{}) Logger
 	// WithField return copy of Logger with predefined single field.
 	WithField(string, interface{}) Logger
+
+	// Copy copies logger
+	Copy() Logger
```
ruz: This is wrong. You don't need copy if thing is immutable and all methods do CopyOnWrite

bigbes: Removed Copy, Set -> With

insolar/message/parcel.go
bigbes: nil'able -> means NoLevel

instrumentation/inslogger/logger.go
```diff
@@ -41,6 +42,20 @@ func SetLogger(ctx context.Context, l insolar.Logger) context.Context {
 	return context.WithValue(ctx, loggerKey{}, l)
 }
 
+// SetLoggerLevel returns context with provided insolar.LogLevel and set logLevel on logger,
+func WithLoggerLevel(ctx context.Context, logLevel insolar.LogLevel) context.Context {
+	if logLevel != insolar.NoLevel {
+		oldLogger := FromContext(ctx)
+		logCopy := oldLogger.Copy()
+		if err := logCopy.SetLevelNumber(logLevel); err != nil {
```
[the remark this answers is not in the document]
bigbes: done

instrumentation/inslogger/logger.go
```diff
@@ -37,14 +38,29 @@ func FromContext(ctx context.Context) insolar.Logger {
...
 }
 
+// SetLoggerLevel returns context with provided insolar.LogLevel and set logLevel on logger,
+func WithLoggerLevel(ctx context.Context, logLevel insolar.LogLevel) context.Context {
+	if logLevel != insolar.NoLevel {
+		logger := FromContext(ctx)
+		if loggerCopy, err := logger.WithLevelNumber(logLevel); err != nil {
+			logger.Error("failed to set log level: ", err.Error())
+			return ctx
+		} else {
```
golangcibot: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary) (from `golint`)

api/call.go
```diff
@@ -36,11 +36,12 @@ import (
...
-	Method    string `json:"method"`
-	Params    []byte `json:"params"`
-	Seed      []byte `json:"seed"`
-	Signature []byte `json:"signature"`
+	Reference string      `json:"reference"`
+	Method    string      `json:"method"`
+	Params    []byte      `json:"params"`
+	Seed      []byte      `json:"seed"`
+	Signature []byte      `json:"signature"`
+	LogLevel  interface{} `json:"logLevel"`
```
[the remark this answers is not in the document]
bigbes: done

PR #1455, opened 2019-03-19
INS-2136: pulsewatcher json formatting support

afiskon:


cmd/pulsewatcher/pulsewatcher.go
```diff
@@ -65,9 +65,104 @@ func moveBack(reader io.Reader) {
...
+
+func parseInt64(str string) int64 {
+	res, err := strconv.ParseInt(str, 10, 64)
+	if err != nil {
+		res = -1
+	}
+	return res
+}
+
+func displayResultsJson(results [][]string, buffer *bytes.Buffer, notReady bool) {
```
golangcibot: func displayResultsJson should be displayResultsJSON (from `golint`)

cmd/pulsewatcher/pulsewatcher.go
```diff
@@ -65,9 +65,104 @@ func moveBack(reader io.Reader) {
...
+	if err != nil {
+		panic(err) // should never happen
+	}
+	fmt.Print(string(jsonDoc))
+	fmt.Print("\n\n")
+}
+
 func main() {
 	var configFile string
+	var useJsonFormat bool
```
golangcibot: var `useJsonFormat` should be `useJSONFormat` (from `golint`)

cmd/pulsewatcher/pulsewatcher.go
```diff
@@ -65,9 +65,104 @@ func moveBack(reader io.Reader) {
...
+		PulseNumber int64
+		ActiveListSize int64
+		WorkingListSize int64
+		Role string
+		Error string
+	}
+
+	doc := make([]DocumentItem, len(results))
+
+	for i, res := range results {
```
aalekseev-ru: use mapping from not named fields too named fields can cause more trouble in future

cmd/pulsewatcher/pulsewatcher.go
```diff
@@ -156,57 +251,16 @@ func main() {
...
-
-		lock.Lock()
-		table.AppendBulk(results)
-		lock.Unlock()
-
-		table.Render()
-
-		fmt.Print(buffer)
+		if singleOutput {
+			break
```
aalekseev-ru: can we move a large piece of code that starts in 'for' in another function, that collects data?

afiskon: done.

cmd/pulsewatcher/pulsewatcher.go
```diff
@@ -65,9 +65,180 @@ func moveBack(reader io.Reader) {
...
+		tablewriter.Colors{},
+		tablewriter.Colors{},
+	)
+
+	table.AppendBulk(results)
+	table.Render()
+	fmt.Print(buffer)
+}
+
+func check(err error) {
```
golangcibot: `check` is unused (from `deadcode`)

cmd/pulsewatcher/pulsewatcher.go
afiskon: Sorry, but this time I'm not convinced that a map has any advantages over an array in this regard, nor that this change has any business value. This representation was chosen because it's quite natural for the default table output. I suggest to keep it this way for now.

cmd/pulsewatcher/pulsewatcher.go
afiskon: thanks, fixed!

PR #1464, opened 2019-03-20
License update

ita-sammann:
Updated network license to the actual version.
Switched to single line comments to prevent Goland from breaking build tags.


paras-lehana:
Keep rocking!

PR #1479, opened 2019-03-25
INS-2151: Create protobuf compatibility layer

bigbes:
* Move VirtualRecord/MaterialRecord to record module


insolar/record.go
```diff
@@ -198,6 +195,47 @@ func (ref *Reference) MarshalJSON() ([]byte, error) {
...
+	}
+	copy(ref[:], data)
+	return nil
+}
+func (ref *Reference) UnmarshalJSON(data []byte) error {
+	return json.Unmarshal(data, ref)
+}
+func (ref *Reference) Size() int { return RecordRefSize }
+
+func (ref ID) Marshal() ([]byte, error) { return ref[:], nil }
```
golangcibot: receiver name ref should be consistent with previous receiver name id for ID (from `golint`)

insolar/record.go
```diff
@@ -198,6 +195,47 @@ func (ref *Reference) MarshalJSON() ([]byte, error) {
...
+func (ref *ID) MarshalTo(data []byte) (int, error) {
```
golangcibot: receiver name ref should be consistent with previous receiver name id for ID (from `golint`)

insolar/record.go
```diff
@@ -198,6 +195,47 @@ func (ref *Reference) MarshalJSON() ([]byte, error) {
...
+	copy(data, ref[:])
+	return RecordIDSize, nil
+}
+func (ref *ID) Unmarshal(data []byte) error {
```
golangcibot: receiver name ref should be consistent with previous receiver name id for ID (from `golint`)

insolar/record.go
```diff
@@ -76,11 +76,8 @@ func (id *ID) Hash() []byte {
 }
 
 // Equal checks if reference points to the same record.
-func (id *ID) Equal(other *ID) bool {
```
andreyromancev: Why? It breaks business logic.

insolar/record/record.go
```diff
@@ -0,0 +1,246 @@
...
+	"github.com/insolar/insolar/insolar"
+	"io"
+	"reflect"
+
+	"github.com/pkg/errors"
+)
+
+type SerializableRecord interface {
+	MarshalRecord() ([]byte, error)
+	MarshalRecordTo(data []byte) (int, error)
```
andreyromancev: Do we really need this method?

insolar/record/record.pb.go
```diff
@@ -0,0 +1,5180 @@
...
+var xxx_messageInfo_Record proto.InternalMessageInfo
+
+type isRecord_Union interface {
+	isRecord_Union()
+	Equal(interface{}) bool
+	MarshalTo([]byte) (int, error)
+	Size() int
+}
+
+type Record_Genesis struct {
```
andreyromancev: Can we control these names?

Gopkg.toml
```diff
@@ -24,6 +24,11 @@
 #   go-tests = true
 #   unused-packages = true
 
+required = [
+    "github.com/gogo/protobuf/protoc-gen-gogoslick",
```
AndreyBronin: Fix version of `github.com/gogo/protobuf/protoc-gen-gogoslick` in `install-build-tools` section Makefile

insolar/record/record.pb.go
bigbes: We can, but we don't need to. It won't be used outside "compatibility layer".

insolar/record/record.go
bigbes: It allows us to encode message into existing buffer, and will give us ability to "control" maximum message output size.

insolar/record.go
bigbes: It's something that is demanded by `gogo` marshaller. It passes ID as object, not pointer.
I've found every place in code and changed it appropriately.

Gopkg.toml
bigbes: done

PR #1488, opened 2019-03-26
Ins 1711 refactor blob storage

nordicdyno:
**- What I did**

- implement persistent blob storage
- refactor test, check both storage realization against each other


ledger/storage/blob/db.go
```diff
@@ -0,0 +1,111 @@
...
+
+func (k *dbKey) ID() []byte {
+	return k.id[:]
+}
+
+// ForID returns Blob for provided id.
+func (s *StorageDB) ForID(ctx context.Context, id insolar.ID) (Blob, error) {
+	b, err := s.db.Get(&dbKey{id: id})
+	if err != nil {
+		if err == db.ErrNotFound {
```
egorikas: It mustn't be db.ErrNotFound. It should be blob.ErrNotFound

PR #1498, opened 2019-03-27
NOISSUE: segregate testing port ranges

Vany:
According to environment variable TBN we now choose range of testing ports. This made for reduce port collisions on building platform.

### DIFF
--- a/network/servicenetwork/suite_test.go
+++ b/network/servicenetwork/suite_test.go
@@ -55,6 +55,7 @@ import (
 	"crypto"
 	"fmt"
 	"math/rand"
+	"os"
 	"strconv"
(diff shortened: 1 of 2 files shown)


network/servicenetwork/suite_test.go
```diff
@@ -86,6 +87,21 @@ var (
 	pulseDelta      int32  = 5
 )
 
+func init() {
+	if tbn, ok := os.LookupEnv("TBN"); ok {
```
afiskon: Please rename to TEST_BUILD_NUMBNER

Vany: okay

PR #1520, opened 2019-04-02
INS-2234: test discovery restart fix

krestkrest:
**- What I did**
Fixed flaky TestDiscoveryRestart. Added special network implementation to imitate network delays locally.
**- How to verify it**


network/controller/bootstrap/bootstrap.go
```diff
@@ -221,7 +222,11 @@ func (bc *bootstrapper) Bootstrap(ctx context.Context) (*network.BootstrapResult
 	log.Info("Bootstrapping to discovery node")
 	ctx, span := instracer.StartSpan(ctx, "Bootstrapper.Bootstrap")
 	defer span.End()
-	ch := bc.getDiscoveryNodesChannel(ctx, bc.Certificate.GetDiscoveryNodes(), 1)
+	discoveryNodes := bc.Certificate.GetDiscoveryNodes()
+	if utils.OriginIsDiscovery(bc.Certificate) {
+		discoveryNodes = RemoveOrigin(discoveryNodes, bc.NodeKeeper.GetOrigin().ID())
```
dmitry-panchenko: What we should do when `discoveryNodes` becomes empty?

krestkrest: we should fail the bootstrap process, fixed

network/servicenetwork/flaky_network.go
```diff
@@ -49,3 +49,29 @@
...
+func NewFlakyNetwork(origin network.InternalTransport) network.InternalTransport {
+	return &flakyNetwork{InternalTransport: origin}
+}
+
+type flakyNetwork struct {
+	network.InternalTransport
+}
+
+func (fn *flakyNetwork) SendRequestPacket(ctx context.Context, request network.Request, receiver *host.Host) (network.Future, error) {
+	time.Sleep(time.Millisecond * time.Duration(rand.Int()%flakyLimit))
```
golangcibot: G404: Use of weak random number generator (math/rand instead of crypto/rand) (from `gosec`)

PR #1552, opened 2019-04-11
INS-2193. Heavy entrypoint.

andreyromancev:


ledger/ledger.go
```diff
@@ -15,157 +15,3 @@
 //
 
 package ledger
```
aMFeriuSS: Is it empty file now? For what?

PR #1556, opened 2019-04-11
INS-1969: NodeKeeper tests

krestkrest:
**- What I did**
Added tests for NodeKeeper, simplified internal NodeKeeper states
**- How to verify it**


network/interfaces.go
```diff
@@ -220,31 +220,6 @@ type ConsensusInfo interface {
 	IsJoiner() bool
 }
 
-//go:generate minimock -i github.com/insolar/insolar/network.UnsyncList -o ../testutils/network -s _mock.go
```
AndreyBronin: remove mock file

network/nodenetwork/nodekeeper.go
```diff
@@ -106,12 +107,12 @@ func createOrigin(configuration configuration.HostNetwork, certificate insolar.C
 	), nil
 }
 
-func resolveAddress(configuration configuration.HostNetwork) (string, error) {
-	addr, err := net.ResolveTCPAddr("tcp", configuration.Transport.Address)
+func resolveAddress(configuration configuration.Transport) (string, error) {
```
AndreyBronin: I think we should remove  `resolveAddress ` and `createOrigin ` from NodeNetwork, we can get origin from host network

network/interfaces.go
krestkrest: removed

network/nodenetwork/nodekeeper.go
krestkrest: I will create a separate task for this refactoring

PR #1562, opened 2019-04-12
WLT-288: crypto

pol9kov:
**- What I did**
Added crypto for Ethereum
**- How I did it**
Created interface for keys and KeyProcessor, added custom and common crypto. 
**- How to verify it**
Launch with Ethereum crypto keys.


scripts/insolard/genesis.yaml
```diff
@@ -106,11 +106,11 @@ discovery_nodes:
 #    cert_name: "discovery_cert_23.json"
 
 nodes:
-  -
-    host: "127.0.0.1:63836"
-    role: "virtual"
-    keys_file: "scripts/insolard/nodes/1/keys.json"
-    cert_name: "node_cert_1.json"
+#  -
```
egorikas: It should be uncommented. Because this part of the file provides info about a node, that connects to network after a consensus phase.

platformpolicy/crypto.go
```diff
@@ -14,15 +14,13 @@
 // limitations under the License.
 //
 
-package sign
+package platformpolicy
 
-import (
-	"crypto"
+const (
+	COMMON_CRYPTO cryptoStruct = "COMMON_CRYPTO"
```
andreyromancev: Let's use `CamelCase` in variable names.

scripts/insolard/genesis.yaml
pol9kov: done

andreyromancev:
Please, split crypto package in two commits:
1. The original crypto code.
2. Alterations.

platformpolicy/crypto.go
ita-sammann: And please use iota here. Strings are not needed here

krestkrest:
I think it is better to move `platformpolicy/customcrypto/xcrypto/*` to a separate project, e.g. `github.com/insolar/xcrypto` and fetch it in the main Insolar project as a dependency

platformpolicy/keyprocessor.go
```diff
@@ -0,0 +1,32 @@
...
+
+package platformpolicy
+
+import (
+	"github.com/insolar/insolar/insolar"
+	"github.com/insolar/insolar/platformpolicy/commoncrypto"
+	"github.com/insolar/insolar/platformpolicy/customcrypto"
+)
+
+func NewKeyProcessor() insolar.KeyProcessor {
```
dmitry-panchenko: Why you copy pasted two implementations since you could just add parameter (curve) here?

platformpolicy/crypto.go
pol9kov: done

platformpolicy/keyprocessor.go
pol9kov: It is not copy past. They eat different crypto libs. Look at imports.

dmitry-panchenko: Why we cant just use backported lib in both implementations?

PR #1576, opened 2019-04-16
INS-2310: healthcheck HTTP handler

ita-sammann:
**- What I did**
Added health check HTTP GET handle.

**- How I did it**
Checker reads all discovery nodes from certificated and check that _all_ of them are present in Working List.

**- How to verify it**
Start network, check `http://<node ip>:<API port>/healthcheck`, it should return `200 OK`. Kill one node and check again, it should return `500 FAIL`.


api/healthcheck.go
```diff
@@ -0,0 +1,47 @@
...
+	CertificateManager insolar.CertificateManager
+	NodeNetwork        insolar.NodeNetwork
+}
+
+func NewHealthChecker(cm insolar.CertificateManager, nn insolar.NodeNetwork) *HealthChecker {
+	return &HealthChecker{CertificateManager: cm, NodeNetwork: nn}
+}
+
+// CheckHandler is a HTTP handler for health check.
+func (hc *HealthChecker) CheckHandler(w http.ResponseWriter, r *http.Request) {
```
dmitry-panchenko: Could you please add test for this handler?

ita-sammann: Done

api/healthcheck.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/insolar/insolar/insolar"
+)
+
+// HealthChecker allows to check network status of a node.
+type HealthChecker struct {
+	CertificateManager insolar.CertificateManager
+	NodeNetwork        insolar.NodeNetwork
+}
+
+func NewHealthChecker(cm insolar.CertificateManager, nn insolar.NodeNetwork) *HealthChecker {
```
dmitry-panchenko: Exported function should be commented.

PR #1581, opened 2019-04-16
NOISSUE: WrapBusHandle chan fix

ilyapt:
**- What I did**
- chan will close after read it


insolar/flow/handler/handler.go
```diff
@@ -58,14 +59,24 @@ func (h *Handler) WrapBusHandle(ctx context.Context, parcel insolar.Parcel) (ins
 	go func() {
 		f := thread.NewThread(msg, h.controller)
 		err := f.Run(ctx, h.handles.present(msg))
+		defer func() {
+			recover()
```
golangcibot: Error return value of `recover` is not checked (from `errcheck`)

PR #1582, opened 2019-04-16
NOISSUE: *.pb.go format for protobuf generated files

krestkrest:
**- What I did**
Now all Insolar protobuf generated files follow *.pb.go format and this format is added to code coverage ignore profile.


insolar/record/record.go
```diff
@@ -14,7 +14,6 @@
 // limitations under the License.
 //
 
-// /go:generate protoc -I./vendor -I./ --gogoslick_out=./ insolar/record/record.proto
```
ita-sammann: Why abandon go generate?

bigbes: It doesn't work, and not everyone have installed `protoc`, so preferable way is to delete this

PR #1617, opened 2019-04-19
INS-2344: GetChildren, Flow refactoring

afiskon:


ledger/artifactmanager/handler.go
```diff
@@ -833,6 +720,29 @@ func (h *MessageHandler) saveIndexFromHeavy(
 // 	return rep, nil
 // }
 
+func (h *MessageHandler) saveCodeFromHeavy(
```
golangcibot: U1000: func `(*MessageHandler).saveCodeFromHeavy` is unused (from `unused`)

afiskon: OK, will fix

ledger/handle/get_children_test.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/insolar/insolar/insolar/flow/bus"
+	"github.com/insolar/insolar/insolar/jet"
+	"github.com/insolar/insolar/insolar/message"
+	"github.com/insolar/insolar/ledger/proc"
+	"github.com/insolar/insolar/ledger/recentstorage"
+	"github.com/insolar/insolar/ledger/storage"
+	"github.com/insolar/insolar/ledger/storage/object"
+	"github.com/insolar/insolar/testutils"
+)
+
```
aalekseev-ru: i thought that old test will still working and it's a sign, that nothing was broken

ledger/handle/get_children.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	if err := f.Handle(ctx, jet.Present); err != nil {
+		return err
+	}
+
+	p := s.dep.GetChildren(&proc.GetChildren{
+		Jet:     insolar.ID(jet.Res.Jet),
+		Message: s.Message,
+	})
+	if err := f.Procedure(ctx, p); err != nil {
```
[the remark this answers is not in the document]
afiskon: Nope. We discussed this with Denis and Andrey. When procedure cancels maybe the response is not needed anymore on calling side since there is a canceled procedure on the calling side as well. Thus if necessary the calling side may retry, but we don't do it here.

ledger/handle/get_children_test.go
afiskon: Most of all tests still work. However unit tests were relaying on the method `handleGetChildren` that doesn't exists anymore thus I had to rewrite these unit tests.

ledger/artifactmanager/handler.go
```diff
@@ -120,6 +120,18 @@ func NewMessageHandler(conf *configuration.Ledger) *MessageHandler {
...
+		GetChildren: func(p *proc.GetChildren) *proc.GetChildren {
+			p.Dep.RecentStorageProvider = h.RecentStorageProvider
+			p.Dep.IDLocker = h.IDLocker
+			p.Dep.IndexAccessor = h.IndexAccessor
+			p.Dep.JetCoordinator = h.JetCoordinator
+			p.Dep.JetStorage = h.JetStorage
+			p.Dep.DelegationTokenFactory = h.DelegationTokenFactory
+			p.Dep.RecordAccessor = h.RecordAccessor
+			p.Dep.TreeUpdater = h.jetTreeUpdater
+			p.Dep.IndexSaver = h.IndexSaver
```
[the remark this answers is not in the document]
afiskon: I just followed the code style in this particular file, see lines 95-122. Andrey confirmed that DI is done like this. If necessary we could refactor this but I don't believe we will do it in this task.

ledger/handle/get_children.go
```diff
@@ -0,0 +1,63 @@
...
+		return err
+	}
+
+	rep := &proc.ReturnReply{
+		ReplyTo: s.Message.ReplyTo,
+		Reply:   p.Result.Reply,
+	}
+	if err := f.Procedure(ctx, rep); err != nil {
```
ruz: this code should be just "return f.Procedure(ctx, rep)" as it is what exactly it does.

ledger/handle/get_children_test.go
```diff
@@ -0,0 +1,166 @@
...
+var (
+	domainID = *genRandomID(0)
+)
+
+func genRandomID(pulse insolar.PulseNumber) *insolar.ID {
```
ruz: we have gen.ID() for this. code duplication.

ledger/handle/get_children.go
afiskon: OK, agree.

ledger/handle/get_children_test.go
```diff
@@ -0,0 +1,166 @@
...
+		panic(err)
+	}
+	return insolar.NewID(pulse, buff[:])
+}
+
+func genRefWithID(id *insolar.ID) *insolar.Reference {
+	return insolar.NewReference(domainID, *id)
+}
+
+func genRandomRef(pulse insolar.PulseNumber) *insolar.Reference {
```
ruz: we have generators for almost everything in here, so it should be either re-used or placed into gen package we have for such things.

insolar/gen/reference.go
```diff
@@ -28,6 +28,11 @@ func ID() (id insolar.ID) {
 	return
 }
 
+func IDPointer() *insolar.ID {
```
andreyromancev: There is no point to making a function for that.

ledger/handle/get_children_test.go
afiskon: OK, tnx, fixed.

insolar/gen/reference.go
```diff
@@ -50,3 +55,8 @@ func Reference() (ref insolar.Reference) {
 	fuzz.New().NilChance(0).Fuzz(&ref)
 	return
 }
+
+func ReferencePointer() *insolar.Reference {
```
andreyromancev: There is no point to making a function for that.

ledger/storage/object/indexsaver.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/insolar/insolar/insolar"
+	"github.com/insolar/insolar/insolar/message"
+	"github.com/insolar/insolar/insolar/reply"
+	"github.com/pkg/errors"
+)
+
+//go:generate minimock -i github.com/insolar/insolar/ledger/storage/object.IndexSaver -o ./ -s _mock.go
+
+type IndexSaver interface {
```
andreyromancev: We don't need this component. We have `proc.GetIndex`.

ledger/storage/object/indexsaver.go
```diff
@@ -0,0 +1,60 @@
...
+	SaveIndexFromHeavy(
+		ctx context.Context, jetID insolar.ID, obj insolar.Reference, heavy *insolar.Reference,
+	) (Lifeline, error)
+}
+
+type indexSaverImpl struct {
```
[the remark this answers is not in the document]
afiskon: Tnx, fixed.

insolar/gen/reference.go
afiskon: Agree. Fixed.

insolar/gen/reference.go
afiskon: OK, agree.

ledger/storage/object/indexsaver.go
afiskon: I'm not 100% convinced that rewriting this code to proc.GetIndex Procedure in this task is a good idea. It will affects other message handlers that currently use IndexSaver.SaveIndexFromHeavy (ex saveIndexFromHeavy). We can create a task for this refactoring though if you like.

afiskon:
The patch is outdated. Also after recent discussion with the architect we realized we need a support of synchronous (non interruptible) call from Flow.

PR #1636, opened 2019-04-22
INS-1940: New network alert

ooki2day:
Added a change network claim. This claim gives an address to host with a larger network size. If this claim approved by consensus each node will have to connect to a new host.


afiskon:
Please resolve the conflicts and re-submit PR.

PR #1656, opened 2019-04-24
NOISSUE: fix ci hang

afiskon:


afiskon:
Seemed to help according to https://github.com/insolar/insolar/pull/1655

PR #1670, opened 2019-04-25
WLT-356: heavy sequence record index

uhzxl:
**- What I did**
Added the ability to retrieve records from storage on HM node in an ordered form starting from a certain position
**- How I did it**
Added three DB key scopes: one to save values first and top synced pulse, one to save some pulse meta information (What is the current/next pulse? How much records?) and yet one to save record position and reference on the record itself. Made call to save that index when the HM node receives a message with records from LME.
**- How to verify it**
Run unit tests
Added record sequence index on HM.


ledger/heavy/handler/handler.go
```diff
@@ -43,6 +43,7 @@ type Handler struct {
 	IndexAccessor  object.IndexAccessor
 	IndexModifier  object.IndexModifier
 	DropModifier   drop.Modifier
+	recordSequence object.SequenceRecordModifier
```
egorikas: Where is being inited?

ledger/object/sequence.go
```diff
@@ -0,0 +1,391 @@
...
+		(pos.PulseNumber == other.PulseNumber && pos.Index > other.Index)
+}
+
+type SequenceRecord struct {
+	Position SequenceRecordPosition
+	ID       insolar.ID
+	Record   record.MaterialRecord
+}
+
+type SequenceRecordAccessor interface {
```
egorikas: Add comments to the interfaces too, please

ledger/object/sequence.go
```diff
@@ -0,0 +1,391 @@
...
+type SequenceRecordStorage interface {
+	SequenceRecordAccessor
+	SequenceRecordModifier
+	SequenceRecordCursor
+
+	// UpdatePulse makes "upsert" for top pulse number and sets at old pulse descriptor link to next pulse
+	UpdatePulse(ctx context.Context, pn insolar.PulseNumber) error
+}
+
+type SequenceRecordMemory struct{}
```
egorikas: Struct is exported. Add comments please

ledger/object/sequence.go
```diff
@@ -0,0 +1,391 @@
...
+// ForPosition returns sequence record for provided record position.
+func (s *SequenceRecordDB) ForPosition(ctx context.Context, pos SequenceRecordPosition) (SequenceRecord, error) {
+	s.lock.RLock()
+	defer s.lock.RUnlock()
+
+	return s.get(ctx, pos)
+}
+
+// FromPosition returns sequence record stream from provided record position with corresponding limit.
+func (s *SequenceRecordDB) FromPosition(ctx context.Context, pos SequenceRecordPosition, limit uint32) chan interface{} {
```
egorikas: I think, It's a good idea to return empty chan of interfaces with two different values.
It'd be better to return two channels, with exact result and error

egorikas:
And I suppose, It'd be great to write some unit-tests for the functional at least

ledger/object/sequence.go
uhzxl: Okay, but I wanted to do something similar to how it works in java Reactive Streams (projectreactor.io) when data and errors occurring during the processing of a stream are transmitted through a single stream. What are the pitfalls here?

egorikas: It's not  java. Working through empty interfaces isn't a great thing in go. https://go-proverbs.github.io/

In our experience, casting of empty interfaces is an infinite root of problems

andreyromancev: Using empty interfaces is generally frowned upon in Go community. Let's consider other options.

ledger/heavy/handler/handler.go
uhzxl: Added init

ledger/object/sequence.go
uhzxl: Added

ledger/object/sequence.go
uhzxl: Added

ledger/object/sequence.go
uhzxl: Okay, I replaced it with a simple slice and error value

ledger/object/sequence.go
```diff
@@ -0,0 +1,400 @@
...
+
+func (k topKey) Scope() store.Scope {
+	return store.ScopeReplicaPosition
+}
+
+func (k topKey) ID() []byte {
+	return []byte{byte(k)}
+}
+
+type pulseLabel struct {
```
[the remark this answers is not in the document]
andreyromancev: Actually, this struct is not required at all, if we implement iteration on DB component.

ledger/object/sequence.go
```diff
@@ -0,0 +1,400 @@
...
+
+func (k positionKey) ID() []byte {
+	return append(SequenceRecordPosition(k).PulseNumber.Bytes(),
+		utils.UInt32ToBytes(SequenceRecordPosition(k).Index)...)
+}
+
+func (s *SequenceRecordDB) topPulse() (insolar.PulseNumber, error) {
+	key := topKey(42)
+	buff, err := s.db.Get(key)
+	if err == ErrNotFound {
```
andreyromancev: `err == store.ErrNotFound`

ledger/object/sequence.go
```diff
@@ -0,0 +1,400 @@
...
+	return store.ScopeSequenceRecord
+}
+
+func (k positionKey) ID() []byte {
+	return append(SequenceRecordPosition(k).PulseNumber.Bytes(),
+		utils.UInt32ToBytes(SequenceRecordPosition(k).Index)...)
+}
+
+func (s *SequenceRecordDB) topPulse() (insolar.PulseNumber, error) {
+	key := topKey(42)
```
andreyromancev: Let's not commit unfinished methods.

ledger/object/sequence.go
```diff
@@ -0,0 +1,400 @@
...
+	buff, err := s.db.Get(key)
+	if err == ErrNotFound {
+		return 0, ErrNoSequenceSyncPulses
+	} else if err != nil {
+		return 0, err
+	}
+	return insolar.PulseNumber(binary.BigEndian.Uint32(buff)), nil
```
andreyromancev: `insolar.NewPulseNumber(buff)`

ledger/object/sequence.go
```diff
@@ -0,0 +1,400 @@
...
+	return s.get(ctx, pos)
+}
+
+// FromPosition returns sequence record slice from provided record position with corresponding limit.
+func (s *SequenceRecordDB) FromPosition(ctx context.Context, pos SequenceRecordPosition, limit uint32) ([]SequenceRecord, error) {
+	s.lock.RLock()
+	defer s.lock.RUnlock()
+
+	out := []SequenceRecord{}
+	capacity := uint32(0)
```
andreyromancev: This can be replaced by `len(out)`

ledger/object/sequence.go
```diff
@@ -0,0 +1,400 @@
...
+		if err != nil {
+			return nil, errors.Wrap(err, "[SequenceRecordIndex] failed to get pulse meta")
+		}
+		for ; pos.Index <= curr.topIndex && capacity < limit; pos.Index++ {
+			rec, err := s.get(ctx, pos)
+			if err != nil {
+				return nil, errors.Wrap(err, "[SequenceRecordIndex] failed to get record")
+			}
+			out = append(out, rec)
+			capacity++
```
andreyromancev: We don't check the limit here. We can overflow it and the `capacity == limit` check will fail and iterate to the end.

ledger/object/sequence.go
```diff
@@ -0,0 +1,400 @@
...
+	return out, nil
+}
+
+// UpdatePulse makes "upsert" for top pulse number and sets at old pulse descriptor link to next pulse
+func (s *SequenceRecordDB) UpdatePulse(ctx context.Context, pn insolar.PulseNumber) error {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	prev, err := s.topPulse()
+	if err == ErrNotFound {
```
andreyromancev: `ErrNoSequenceSyncPulses` is returned from `topPulse`, not `ErrNotFound`.
I propose to remove `ErrNoSequenceSyncPulses` completely and use `ErrNotFound` instead.

ledger/object/sequence.go
```diff
@@ -0,0 +1,400 @@
...
+		}
+		return s.updateTop(pn)
+	} else if err != nil {
+		return errors.Wrap(err, "[SequenceRecordIndex] failed to get top pulse number")
+	}
+	pd, err := s.exactPulse(prev)
+	if err != nil {
+		return errors.Wrap(err, "[SequenceRecordIndex] failed to get previous pulse descriptor")
+	}
+	pd.next = pn
```
andreyromancev: `pd` is not written in db. Changes will be lost.

ledger/object/sequence.go
uhzxl: Maybe for testing purposes?

afiskon:
@uhzxl Please resolve the conflicts, make changes named during code review and re-submit PR.

PR #1681, opened 2019-04-25
Ins 2420 logger fix skip call number

nordicdyno:


instrumentation/inslogger/inslogger.go
```diff
@@ -78,11 +78,16 @@ func ContextWithTrace(ctx context.Context, traceid string) context.Context {
 }
 
 func getLogger(ctx context.Context) insolar.Logger {
-	l := ctx.Value(loggerKey{})
-	if l == nil {
-		return logger.GlobalLogger
+	val := ctx.Value(loggerKey{})
+	if val == nil {
+		return logger.GlobalLogger.ChangeSkipFrameCount(1).WithField("loginstance", "global_inslog")
```
ruz: this "loginstance" field is for debugging and should be removed.

insolar/logger.go
```diff
@@ -86,8 +86,8 @@ type Logger interface {
 
 	// WithCaller switch on/off 'caller' field computation.
 	WithCaller(flag bool) Logger
-	// WithSkipFrameCount configures skipFrameCount for 'caller' field computation.
-	WithSkipFrameCount(skipFrameCount int) Logger
+	// ChangeSkipFrameCount changes skipFrameCount by delta value.
+	ChangeSkipFrameCount(delta int) Logger
```
ruz: I don't think this should be renamed. I understand that this method adds or removes delta and it's ok. I prefer "With" prefix, it's more clear that it is a modifier.

ruz: May be better name would be "WithSkipFrameDelta". What do you think?

nordicdyno: 'delta' is already in the function signature, so my solution is just to stop on WithSkipFrameCount with improved godoc description

PR #1698, opened 2019-05-13
Ins 2432 launchnet add common config defaults

nordicdyno:


configuration/configuration.go
```diff
@@ -134,12 +135,36 @@ func NewHolder() *Holder {
...
+		file = file[:len(file)-len(filepath.Ext(file))]
+
+		holder.viper.AddConfigPath(dir)
+		holder.viper.SetConfigName(file)
+	}
+
+	return holder.defaults()
+}
+
+func (h *Holder) defaults() *Holder {
```
golangcibot: receiver name h should be consistent with previous receiver name c for Holder (from `golint`)

PR #1710, opened 2019-05-14
INS-2437: Move TypeGetJet message handling to flow-powered handler/procedure

bigbes:


ledger/light/proc/get_jet.go
```diff
@@ -0,0 +1,38 @@
...
+		replyTo: rep,
+	}
+}
+
+func (p *GetJet) Proceed(ctx context.Context) error {
+	p.replyTo <- p.reply(ctx)
+	return nil
+}
+
+func (p *GetJet) reply(ctx context.Context) bus.Reply {
```
andreyromancev: No need to make a separate function.

bigbes: fixed

PR #1726, opened 2019-05-16
Ins 2463 one call message

ruz:


insolar/message/logicrunner.go
```diff
@@ -117,92 +70,76 @@ func (rr *ReturnResults) AllowedSenderObjectAndRole() (*insolar.Reference, insol
...
+func (*CallMethod) DefaultRole() insolar.DynamicRole {
 	return insolar.DynamicRoleVirtualExecutor
 }
 
 // DefaultTarget returns of target of this event.
-func (cc *CallConstructor) DefaultTarget() *insolar.Reference {
-	if cc.SaveAs == Delegate {
-		return &cc.ParentRef
+func (cm *CallMethod) DefaultTarget() *insolar.Reference {
+	if cm.CallType == CTSaveAsChild {
```
golangcibot: ifElseChain: rewrite if-else to switch statement (from `gocritic`)

insolar/message/logicrunner.go
```diff
@@ -117,92 +70,76 @@ func (rr *ReturnResults) AllowedSenderObjectAndRole() (*insolar.Reference, insol
...
 	}
-	return genRequest(cc.PulseNum, MustSerializeBytes(cc), &insolar.DomainID)
 }
 
-func (cc *CallConstructor) GetReference() insolar.Reference {
-	return *genRequest(cc.PulseNum, MustSerializeBytes(cc), &insolar.DomainID)
+func (cm *CallMethod) GetReference() insolar.Reference {
+	if cm.CallType == CTMethod {
+		return *cm.Object
+	} else {
```
golangcibot: `if` block ends with a `return` statement, so drop this `else` and outdent its block (from `golint`)

api/contract.go
```diff
@@ -119,10 +119,12 @@ func (s *ContractService) CallConstructor(r *http.Request, args *CallConstructor
 		return errors.Wrap(err, "can't get domain reference")
 	}
 
+	base := testutils.RandomRef()
+
 	contractID, err := s.runner.ArtifactManager.RegisterRequest(
 		ctx,
 		insolar.GenesisRecord.Ref(),
-		&message.Parcel{Msg: &message.CallConstructor{PrototypeRef: testutils.RandomRef()}}, // TODO protoRef?
+		&message.Parcel{Msg: &message.CallMethod{Base: &base}}, // TODO protoRef?
```
aalekseev-ru: maybe remove TODO?

contractrequester/contractrequester.go
```diff
@@ -169,98 +156,40 @@ func (cr *ContractRequester) CallMethod(ctx context.Context, base insolar.Messag
...
-		cr.Sequence++
-		seq = cr.Sequence
-		msg.Sequence = seq
-		ch = make(chan *message.ReturnResults, 1)
-		cr.ResultMap[seq] = ch
-
-		cr.ResultMutex.Unlock()
-	}
+func (cr *ContractRequester) CallMethod(ctx context.Context, inMsg insolar.Message) (insolar.Reply, error) {
+	return cr.Call(ctx, inMsg)
```
aalekseev-ru: you check type of returned message in CallConstructor but here you don't, it's strange. What if we get reply.CallConstructor ?

api/contract.go
ita-sammann: Then we should not use of RandomRef()

contractrequester/contractrequester.go
ruz: also, Call may return other types of replies and pretty much valid, for example RegisterReply

PR #1742, opened 2019-05-20
INS-2438 type get pending request id to flow

aalekseev-ru:
**- What I did**
move typeGetPendingRequest to flow
**- How I did it**
add handler, add proc, rewrite register function, add tests
**- How to verify it**
run tests


insolar/reply/ledger.go
```diff
@@ -109,6 +109,17 @@ func (e *HasPendingRequests) Type() insolar.ReplyType {
 	return TypePendingRequests
 }
 
+// HasPendingRequests contains unclosed requests for an object.
+type PendingRequest struct {
```
afiskon: The comment doesn't match the structure.

ledger/light/artifactmanager/handler.go
```diff
@@ -270,7 +270,7 @@ func (h *MessageHandler) setHandlersForLight(m *middleware) {
 	h.Bus.MustRegister(
 		insolar.TypeGetPendingRequestID,
 		BuildMiddleware(
-			h.handleGetPendingRequestID,
+			h.FlowDispatcher.WrapBusHandle,
```
afiskon: This piece of code looks funny. I don't believe we keep middlewares during migrations to the flow. The handler has to contain the code of checkJet.

afiskon: In fact IIRC getFetchJet already does this thus you can get rid of middlewares here.

PR #1746, opened 2019-05-21
INS-1635: CM: don't call Stop() in Start() is not called yet

ivanshibitov:
**- What I did**
Prevent call Stop() before Start() on component manager

**- How I did it**

**- How to verify it**

### DIFF
--- a/component/manager.go
+++ b/component/manager.go
@@ -194,7 +194,7 @@ func (m *Manager) Stop(ctx context.Context) error {
 	m.startStopLock.Lock()
 	defer m.startStopLock.Unlock()
 
-	if m.started == false {
+	if m.started {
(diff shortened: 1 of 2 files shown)


component/manager.go
```diff
@@ -183,6 +191,13 @@ func (m *Manager) GracefulStop(ctx context.Context) error {
 
 // Stop invokes Stop method of all components which implements Starter interface
 func (m *Manager) Stop(ctx context.Context) error {
+	m.startStopLock.Lock()
+	defer m.startStopLock.Unlock()
+
+	if m.started == false {
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `!m.started` (from `gosimple`)

PR #1751, opened 2019-05-21
INS-2484: drop not used arguments in GetObject

ruz:
we don't use these and don't plan to use soon


insolar/message/ledger.go
```diff
@@ -90,7 +90,6 @@ type GetObject struct {
 	ledgerMessage
 	Head     insolar.Reference
 	State    *insolar.ID // If nil, will fetch the latest state.
```
andreyromancev: Remove `State` as well, please.

ruz: are you sure?

PR #1752, opened 2019-05-21
NOISSUE: fix global log configuration

nordicdyno:


configuration/configuration.go
```diff
@@ -89,6 +89,9 @@ func (h *Holder) Init(required bool) (*Holder, error) {
 		if _, ok := err.(viper.ConfigFileNotFoundError); !ok {
 			return nil, err
 		}
+		// read env vars if config file is not required and viper failed to load it.
+		h.viper.AutomaticEnv()
+		h.viper.Unmarshal(&h.Configuration)
```
golangcibot: Error return value of `h.viper.Unmarshal` is not checked (from `errcheck`)

log/log_test.go
```diff
@@ -39,6 +42,71 @@ func capture(f func()) string {
...
+
+	if val == "debug" {
+		assert.Containsf(t,
+			capture(func() { Debug("HelloWorld") }),
+			"HelloWorld", "Debug should work on level %v", val)
+	} else {
+		assert.NotContainsf(t, capture(func() { Debug("HelloWorld") }),
+			"HelloWorld", "Debug should not work on level %v", val)
+	}
+	// assert.NotContains(t, capture(func() { Debug("HelloWorld") }),
```
ivanshibitov: Do we need it here?

log/log_test.go
```diff
@@ -39,6 +42,71 @@ func capture(f func()) string {
 	return buf.String()
 }
 
+var logLevelEnvVarName = "INSOLAR_LOG_LEVEL"
+
+func testWithEnvVar(t *testing.T) {
+	val := strings.ToLower(os.Getenv(logLevelEnvVarName))
+	fmt.Printf("testWithEnvVar: %v=%v\n", logLevelEnvVarName, val)
+
+	assert.Containsf(t,
```
ivanshibitov: Do we need here exactly 'assert'? Might be it's better to use 'require'

nordicdyno: it's ok, the test still fails with assert

PR #1753, opened 2019-05-22
INS-2487: Added error handler

pol9kov:
**- What I did**
Added error handler 

**- How I did it**
Added error handler 

**- How to verify it**
Return error to RegisterNode method in NodeDomain contract

### DIFF
--- a/cmd/insolar/certgen.go
+++ b/cmd/insolar/certgen.go
@@ -60,12 +60,12 @@ func extractReference(response []byte, requestTypeMsg string) insolar.Reference 
 	r := RegisterResult{}
 	err := json.Unmarshal(response, &r)
 	checkError(fmt.Sprintf("Failed to parse response from '%s' node request", requestTypeMsg), err)
+	if verbose {
+		fmt.Println("Response:", string(response))
+	}
 	if r.Error != "" {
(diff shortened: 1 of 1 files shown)


cmd/insolar/certgen.go
```diff
@@ -59,6 +60,10 @@ func extractReference(response []byte, requestTypeMsg string) insolar.Reference
 	r := RegisterResult{}
 	err := json.Unmarshal(response, &r)
 	checkError(fmt.Sprintf("Failed to parse response from '%s' node request", requestTypeMsg), err)
+	if r.Error != "" {
```
ivanshibitov: Use checkError here

pol9kov: checkError method signature doesn't fit

cmd/insolar/certgen.go
```diff
@@ -59,6 +60,10 @@ func extractReference(response []byte, requestTypeMsg string) insolar.Reference
...
+		fmt.Printf("Error while '%s' occured : %s \n", requestTypeMsg, r.Error)
+		os.Exit(1)
+	}
 	if verbose {
```
ivanshibitov: I think it should be before your check

cmd/insolar/certgen.go
ivanshibitov: ok

PR #1768, opened 2019-05-23
request should be stored next to object it belongs to

ruz:
all pending requests should be tracked by object, requests should be
executed in order and object should be able to tell which one goes first
and which is last. results also go next to requests, so next to objects.

### DIFF
--- a/logicrunner/artifacts/client.go
+++ b/logicrunner/artifacts/client.go
@@ -92,7 +92,7 @@ func (m *client) RegisterRequest(
 	switch request.CallType {
 	case record.CTMethod:
 		recRef = request.Object
-	case record.CTSaveAsChild, record.CTSaveAsDelegate:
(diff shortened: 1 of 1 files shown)


logicrunner/artifacts/client.go
```diff
@@ -87,9 +87,18 @@ func (m *client) RegisterRequest(
...
+	var recRef *insolar.Reference
+	switch request.CallType {
+	case record.CTMethod:
+		recRef = request.Object
+	case record.CTSaveAsChild, record.CTSaveAsDelegate, record.CTGenesis:
+		hash := record.HashVirtual(m.PCS.ReferenceHasher(), virtRec)
+		recID := insolar.NewID(currentPN, hash)
+		recRef = insolar.NewReference(insolar.DomainID, *recID)
+	default:
+		return nil, errors.New("not supported call type "+ request.CallType.String())
```
ita-sammann: Doesn't panic suit better here?

PR #1800, opened 2019-05-30
INS-2529: refactor RegisterRequest, use Flow and retries

afiskon:


logicrunner/handle_calls.go
```diff
@@ -30,6 +30,11 @@ import (
 	"go.opencensus.io/trace"
 )
 
+// PSE AGN means "please again" in radio jargon.
```
golangcibot: comment on exported var `ErrRetry` should be of the form `ErrRetry ...` (from `golint`)

afiskon: fixed

aalekseev-ru: i don't think that is a good idea to use jargon in comments/code

logicrunner/handle_calls.go
```diff
@@ -30,6 +30,11 @@ import (
...
+// This text was chosen because it's very unlikely anyone
+// will use the same text for anything else in our code base.
+var ErrRetry = errors.New("PSE AGN")
```
aalekseev-ru: why don't use common message? if we start to use any jargons that we likes we also can start write code/comments in different languages

PR #1806, opened 2019-05-30
INS-2531: Fix linter errors

ita-sammann:
Fixed multiple linter errors


insolar/jet/bits_test.go
```diff
@@ -17,11 +17,60 @@
...
 import (
+	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
 
+// bitsToString converts byte slice to plain string representation in bits.
+func bitsToString(bitslice []byte) string {
```
Dzirtik: Why code has been moved to test?

egorikas: We discussed this method yesterday. It's being used only in tests. Because of the situation, linter shows warnings about `dead code`

PR #1828, opened 2019-06-03
INS-2578: failfast will stop tests on failure preventing timeouts. Th…

LinnikD:


Makefile
```diff
@@ -178,7 +178,7 @@ ci_test_slow:
 
 .PHONY: ci_test_func
 ci_test_func:
-	CGO_ENABLED=1 go test $(TEST_ARGS) -json -tags functest -v ./functest -count 3 | tee ci_test_func.json
+	CGO_ENABLED=1 INSOLAR_LOG_LEVEL=error go test $(TEST_ARGS) -json -tags functest -v ./functest -count 3 -failfast | tee ci_test_func.json
```
Tsovak: I really think the `failfast` parameter does not need here. 
btw. failfast won't solve the problem with hanging

LinnikD: It will. After first failure all nodes fail some requests with timeout error. For now it is known issue, but fix can take time.

PR #1838, opened 2019-06-04
INS-2002: extract some of code from lr.OnPulse into es.OnPulse

ruz:


logicrunner/executionstate.go
```diff
@@ -81,3 +84,78 @@ func (es *ExecutionState) releaseQueue() ([]ExecutionQueueElement, bool) {
...
+			sendExecResults = true
+
+			// TODO: this should return delegation token to continue execution of the pending
+			messages = append(
+				messages,
+				&message.StillExecuting{
+					Reference: ref,
+				},
+			)
+		} else {
```
golangcibot: elseif: can replace 'else {if cond {}}' with 'else if cond {}' (from `gocritic`)

PR #1845, opened 2019-06-05
INS-2451: fix goroutines leak in network integration tests

afiskon:


afiskon:
Nope, this doesn't solve the issue.

PR #1861, opened 2019-06-07
INS-2664: Fix reply error and add flow canceled code.

andreyromancev:


ledger/light/handle/init.go
```diff
@@ -186,3 +194,16 @@ func (s *Init) Past(ctx context.Context, f flow.Flow) error {
 		Err:     errors.New("no past handler"),
 	}, false)
 }
+
+func (s *Init) replyError(ctx context.Context, replyTo *wmessage.Message, err error) {
+	errCode := payload.CodeUnknown
+	switch err { // nolint
+	case flow.ErrCancelled:
```
egorikas: please, make if instead

PR #1864, opened 2019-06-10
INS-2600: timeout in TestSeedManager_Add changed to 10 minutes

ilyapt:
**- What I did**
- changed timeout to 10 minutes

### DIFF
--- a/api/seedmanager/seedmanager_test.go
+++ b/api/seedmanager/seedmanager_test.go
@@ -53,45 +53,45 @@ func getSeed(t *testing.T) Seed {
 }
 
 func TestSeedManager_Add(t *testing.T) {
-	sm := NewSpecified(time.Duration(10 * time.Minute), DefaultCleanPeriod)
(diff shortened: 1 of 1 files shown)


api/seedmanager/seedmanager_test.go
```diff
@@ -53,45 +53,45 @@ func getSeed(t *testing.T) Seed {
...
+	sm.seedPool[seed] = time.Now().UnixNano() - 1000
 	require.False(t, sm.Exists(seed))
 }
 
 func TestRace(t *testing.T) {
 	const numConcurrent = 15
 
-	expTime := time.Duration(2 * time.Millisecond)
-	cleanPeriod := time.Duration(1 * time.Millisecond)
+	expTime := time.Duration(1 * time.Minute)
```
megge-dream: This test checks race between `delete()` call in `Exists()` func and `deleteExpired()` call in goroutine, started in sm constructor. You shouldn't change this test, because we don't care if seed exist.

PR #1870, opened 2019-06-10
Ins 2430 remove GIL from network

Vany:


network/servicenetwork/servicenetwork.go
```diff
@@ -146,15 +151,19 @@ func (n *ServiceNetwork) SetGateway(g network.Gateway) {
 	n.gatewayMu.Unlock()
 	ctx := context.Background()
 	if n.gateway.NeedLockMessageBus() {
```
AndreyBronin: You should write unit test for SetGateway()

network/servicenetwork/servicenetwork_test.go
```diff
@@ -343,3 +341,39 @@ func TestServiceNetwork_processIncoming(t *testing.T) {
 	_, err = serviceNetwork.processIncoming(ctx, data)
 	assert.Error(t, err)
 }
+
+func TestServiceNetwork_SetGateway(t *testing.T) {
```
AndreyBronin: you should also check set to the same state and set to nil state behaviors

insolar/network.go
```diff
@@ -47,6 +47,8 @@ type Network interface {
 	Leave(ctx context.Context, ETA PulseNumber)
 	// GetState returns our current thoughs about whole network
 	GetState() NetworkState
+	// SetOperableFunc registers callback for notifying of network state
+	SetOperableFunc(f func(ctx context.Context, operable bool))
```
krestkrest: If we want to notify other insolar components of network state change, we should pass state instead of a boolean flag that has uncertain nature. The name of the callback also should be picked more carefully

network/servicenetwork/servicenetwork_test.go
Vany: firs case in array is nonetwork to nonetwoek

insolar/network.go
Vany: No. We not need to force users to understand what we allowing in what state.

PR #1888, opened 2019-06-12
Ins 2692 move get pending to proc

ivanshibitov:
**- What I did**
1) launch queueProcessor for AdditionalCallFromPreviousExecutor
2) launch UnsafeGetLedgerPendingRequest in flow

**- How I did it**

**- How to verify it**

### DIFF
--- a/logicrunner/executionstate.go
+++ b/logicrunner/executionstate.go
@@ -56,6 +56,7 @@ func NewExecutionState(ref insolar.Reference) *ExecutionState {
 		Ref:         ref,
 		CurrentList: NewCurrentExecutionList(),
 		Queue:       make([]Transcript, 0),
+		pending:     message.PendingUnknown,
 	}
 }
 
--- a/logicrunner/handle_calls.go
+++ b/logicrunner/handle_calls.go
@@ -227,22 +227,12 @@ func (h *HandleAdditionalCallFromPreviousExecutor) handleActual(
 	os.Unlock()
 
 	es.Lock()
+	if es.pending == message.PendingUnknown {
+		es.pending = message.NotPending
+	}
 	qElement := *NewTranscript(ctx, msg.Parcel, msg.Request, lr.pulse(ctx), es.Ref)
 	es.Queue = append(es.Queue, qElement)
 	es.Unlock()
-
-	procClarifyPendingState := ClarifyPendingState{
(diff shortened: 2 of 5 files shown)


logicrunner/logicrunner_unit_test.go
```diff
@@ -185,27 +185,13 @@ func (suite *LogicRunnerTestSuite) TestHandleAdditionalCallFromPreviousExecutor(
...
-			expectedClarifyPendingStateCtr: 1,
 			expectedStartQueueProcessorCtr: 1,
 		},
-		{
-			name:                           "Both procedures fail",
-			clarifyPendingStateResult:      fmt.Errorf("ClarifyPendingState failed"),
-			startQueueProcessorResult:      fmt.Errorf("StartQueueProcessorIfNeeded failed"),
-			expectedClarifyPendingStateCtr: 1,
-			expectedStartQueueProcessorCtr: 0,
-		},
```
afiskon: Test should be fixed, not removed. Or rewritten according to the changes.

afiskon: Oops never mind. I though it's another test.

logicrunner/executionstate.go
```diff
@@ -56,6 +56,7 @@ func NewExecutionState(ref insolar.Reference) *ExecutionState {
 		Ref:         ref,
 		CurrentList: NewCurrentExecutionList(),
 		Queue:       make([]Transcript, 0),
+		pending:     message.PendingUnknown,
```
ruz: this is default, what the point?

logicrunner/handle_calls.go
```diff
@@ -57,6 +57,7 @@ func (h *HandleCall) sendToNextExecutor(ctx context.Context, es *ExecutionState,
 		ObjectReference: es.Ref,
 		Parcel:          parcel,
 		Request:         request,
+		Pending:         es.pending,
```
ruz: this is wrong. It should be like this:
`
if es.pending == Unknown {
  msg.Pending = NotInPending
} else {
  msg.Pending = es.pending
}
`

logicrunner/logicrunner_unit_test.go
ruz: please return tests back

PR #1901, opened 2019-06-12
NOISSUE MessageBusLocker removed

AndreyBronin:
**- What I did**
remove unused MessageBusLocker interface 

**- How to verify it**
tests passed


Tsovak:
approved by me because the check was pushed with an incorrect name

PR #1914, opened 2019-06-18
INS-2747 zero indexes in hot data

egorikas:
**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/ledger/light/pulsemanager/pulsemanager.go
+++ b/ledger/light/pulsemanager/pulsemanager.go
@@ -266,7 +266,7 @@ func (m *PulseManager) getExecutorHotData(
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


ledger/light/pulsemanager/pulsemanager.go
```diff
@@ -266,7 +266,7 @@ func (m *PulseManager) getExecutorHotData(
 		return nil, err
 	}
 
-	hotIndexes := make([]message.HotIndex, len(bucks))
+	hotIndexes := []message.HotIndex{}
```
ivanshibitov: Might be it's better like that :
hotIndexes := make([]message.HotIndex, 0, len(bucks))
?
To prevent reallocation due to consequence "append"s

PR #1952, opened 2019-06-24
INS-2779: fix bad assertion type in CallHelloWorld test

bigbes:


functest/helloworld_test.go
```diff
@@ -224,8 +224,7 @@ func TestCallHelloWorld(t *testing.T) {
 	r.NoError(err)
 	// tip: right now deduplication is not presented in our system, so count should be
 	//      more or equal to our number of requests
-	a.GreaterOrEqual(100, count)
```
ruz: considering comment I would rather write `a.GreaterOrEqual(count, 100)`

sorry, couldn't ignore inner voice

PR #1964, opened 2019-06-26
Wlt 600 delete ba

pol9kov:
**- What I did**
Fixed the bug with createMember method. Burn addreess wasn't deleted
**- How I did it**
Edited root domain method GetBurnAddress.  
**- How to verify it**
Launch TestCreateMemberWhenNoBurnAddressesLeft.


application/contract/rootdomain/rootdomain.go
```diff
@@ -97,12 +97,15 @@ func (rd *RootDomain) AddBurnAddress(burnAddress string) error {
 	return nil
 }
 
-func (rd RootDomain) GetBurnAddress() (string, error) {
+func (rd *RootDomain) GetBurnAddress() (result string, err error) {
```
megge-dream: Functions and methods should not use named return parameters (our code style, 3.2)

application/contract/rootdomain/rootdomain.go
```diff
@@ -97,12 +97,15 @@ func (rd *RootDomain) AddBurnAddress(burnAddress string) error {
...
 	if len(rd.FreeBurnAddresses) == 0 {
-		return "", fmt.Errorf("no more burn address left")
+		return "", fmt.Errorf("no more burn addresses left")
 	}
 
-	return rd.FreeBurnAddresses[0], nil
+	result = rd.FreeBurnAddresses[0]
+	rd.FreeBurnAddresses = rd.FreeBurnAddresses[1:]
```
megge-dream: What if we get error while adding new member to maps? Should we return burn address to list of free addresses in that case?

pol9kov: Yes. I created rollBack func in this PR https://github.com/insolar/insolar/pull/1966

application/contract/rootdomain/rootdomain.go
pol9kov: In this case it is not get method in standard meaning.

pol9kov: RootDomain doesn't have field burnAddress. Instead in this method we get one burn address from list.

megge-dream: I'm talking about return parameters, not get prefix in method name

PR #1966, opened 2019-06-26
Ins 2774 same pubkey

pol9kov:
**- What I did**
Wrote functest for createMember method, that check case when we create two members with same public keys. Added rollBack function to createMember
**- How I did it**
Wrote functest
**- How to verify it**
Launch TestCreateMembersWithSameName


functest/create_member_test.go
```diff
@@ -47,3 +47,17 @@ func TestCreateMemberWithBadKey(t *testing.T) {
...
 }
+
+func TestCreateMembersWithSameName(t *testing.T) {
+	member, err := newUserWithKeys()
+	require.NoError(t, err)
+	member.ref = root.ref
+
+	addBurnAddresses(t)
+
+	_, err = signedRequest(member, "contract.createMember", map[string]interface{}{})
```
megge-dream: This request must be retried if it was duplicated for some reason

functest/create_member_test.go
```diff
@@ -47,3 +47,18 @@ func TestCreateMemberWithBadKey(t *testing.T) {
...
 }
+
+func TestCreateMembersWithSameName(t *testing.T) {
+	member, err := newUserWithKeys()
+	require.NoError(t, err)
+	member.ref = root.ref
+
+	addBurnAddresses(t)
+
+	_, err = retryableCreateMember(member, "contract.createMember", map[string]interface{}{}, false)
```
megge-dream: Last param in `retryableCreateMember` must be `true` (we want new valid publicKey for every retry attempt)

functest/create_member_test.go
```diff
@@ -47,3 +47,18 @@ func TestCreateMemberWithBadKey(t *testing.T) {
...
+	require.NoError(t, err)
+
+	_, err = retryableCreateMember(member, "contract.createMember", map[string]interface{}{}, false)
```
megge-dream: Here use `signedRequest` because `member for this publicKey already exist` error is that we want to get, no need to retry `createMember` call

api/call.go
```diff
@@ -124,6 +124,10 @@ func processError(err error, extraMsg string, resp *requester.ContractAnswer, in
 	insLog.Error(errors.Wrapf(err, "[ CallHandler ] %s", extraMsg))
 }
 
+func setRequestLogLevel(contractRequest requester.Request) {
```
golangcibot: `setRequestLogLevel` is unused (from `deadcode`)

PR #1968, opened 2019-06-26
INS 2774 removed rootdomain from Call arguments

pol9kov:
**- What I did**
Move rootDomain reference from Call method arguments to member field
**- How I did it**
Removed rootDomain reference from Call method arguments. Added rootDomain reference field to member and edited memberconstructor.
**- How to verify it**
Call call method


bigbes:
Three merges with master? Maybe its simpler to use rebase here?

application/contract/member/member.go
```diff
@@ -160,16 +162,16 @@ func (m *Member) registerNodeCall(rd insolar.Reference, params map[string]interf
 		return nil, fmt.Errorf("incorect input: failed to get 'publicKey' param")
 	}
 
-	return m.registerNode(rd, publicKey, role)
+	return m.registerNode(m.RootDomain, publicKey, role)
```
megge-dream: If `registerNode` is method of `m`, why use `m.RootDomain` as first param? Just get `m.RootDomain` inside `registerNode` func

application/contract/member/member.go
```diff
@@ -118,37 +120,37 @@ func (m *Member) Call(rootDomain insolar.Reference, signedRequest []byte) (inter
...
-func (m *Member) getNodeRefCall(rd insolar.Reference, params map[string]interface{}) (interface{}, error) {
+func (m *Member) getNodeRefCall(params map[string]interface{}) (interface{}, error) {
 
 	publicKey, ok := params["publicKey"].(string)
 	if !ok {
 		return nil, fmt.Errorf("incorect input: failed to get 'publicKey' param")
 	}
 
-	return m.getNodeRef(rd, publicKey)
+	return m.getNodeRef(m.RootDomain, publicKey)
```
megge-dream: get `m.RootDomain` inside `getNodeRef` func too

PR #1982, opened 2019-06-27
INS-2707: declare OutgoingRequest protobuf record

afiskon:


afiskon:
Most of the changes are in record.proto. The rest of the code is auto generated.

PR #1992, opened 2019-07-01
Ins 2828 get jettree from heavy

ivanshibitov:
**- What I did**
1. Go to heavy from JetFetcher
2. handler GetJet message on heavy

**- How I did it**

**- How to verify it**


insolar/jet/fetcher.go
```diff
@@ -301,9 +301,17 @@ func (tu *fetcher) nodesForPulse(ctx context.Context, pulse insolar.PulseNumber)
 
 	res, err := tu.Nodes.InRole(pulse, insolar.StaticRoleLightMaterial)
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "[ nodesForPulse ] Can't get node of 'light' role")
```
andreyromancev: We don't use special formatting in error text.

ledger/heavy/handler/handler_test.go
```diff
@@ -0,0 +1,70 @@
...
+		actual bool
+	}{
+		{name: "Actual", actual: true},
+		{name: "NotActual", actual: false},
+	}
+	for _, tt := range ttable {
+		t.Run(tt.name, func(t *testing.T) {
+			testCtx := inslogger.TestContext(t)
+			testPulseNumber := insolar.PulseNumber(998999)
+			testID := testutils.RandomID()
```
andreyromancev: `gen.ID()`

ledger/heavy/handler/handler_test.go
```diff
@@ -0,0 +1,70 @@
...
+		name   string
+		actual bool
+	}{
+		{name: "Actual", actual: true},
+		{name: "NotActual", actual: false},
+	}
+	for _, tt := range ttable {
+		t.Run(tt.name, func(t *testing.T) {
+			testCtx := inslogger.TestContext(t)
+			testPulseNumber := insolar.PulseNumber(998999)
```
andreyromancev: `testPulseNumber := insolar.FirstPulseNumber + N`

insolar/jet/fetcher.go
```diff
@@ -301,9 +301,17 @@ func (tu *fetcher) nodesForPulse(ctx context.Context, pulse insolar.PulseNumber)
 
 	res, err := tu.Nodes.InRole(pulse, insolar.StaticRoleLightMaterial)
```
andreyromancev: This will return `node.ErrNoNodes`. We need to catch it and **only then** return heavy node for **current pulse**.

andreyromancev: Also, please write a test for this case.

ivanshibitov: I think this logic is not correct. When we have just started on existing db, we will have Light nodes, but all of them might not have actual jets.
About tests - yes, sure!

PR #1995, opened 2019-07-01
INS-2707: Request interface

afiskon:


insolar/record/record.go
```diff
@@ -129,3 +129,24 @@ func (Genesis) GetImage() *insolar.Reference {
 func (Genesis) GetIsPrototype() bool {
 	return false
 }
+
+type Request interface {
+	RequestObject() insolar.Reference
```
andreyromancev: Let's remove `Request` prefix.

afiskon: done

PR #2013, opened 2019-07-03
Ins 2863 outgoings

egorikas:
**- What I did**

**- How I did it**

**- How to verify it**


ledger/light/handle/set_outgoing_request.go
```diff
@@ -0,0 +1,116 @@
...
+	var create = request.CallType == record.CTSaveAsChild || request.CallType == record.CTSaveAsDelegate
+
+	if create {
+		return fmt.Errorf("SetOutgoingRequest can't be a cretion request")
+	}
+
+	return s.setRequest(ctx, msg, request, f)
+}
+
+func (s *SetOutgoingRequest) setRequest(
```
andreyromancev: Let's move this in the `Present`. Since we don't have the second branch, there is no point for a separate funtion.

PR #2021, opened 2019-07-03
sequence removed from request

ilyapt:
**- What I did**
- sequence removed from request
- use hash(request) as key for save to map
- use part of requestID as key for find in map


logicrunner/requestsexecutor.go
```diff
@@ -157,7 +156,7 @@ func (e *requestsExecutor) SendReply(
 		ctx,
 		&message.ReturnResults{
 			Target:   target,
-			Sequence: seq,
+			ReqRef:  *transcript.RequestRef,
```
ruz: Please use longer name for the field, either RequestRef or Request.

contractrequester/contractrequester.go
```diff
@@ -118,16 +119,25 @@ func (cr *ContractRequester) Call(ctx context.Context, inMsg insolar.Message) (i
...
+		if err != nil {
+			return nil, errors.Wrap(err, "[ ContractRequester::Call ] Failed to marshal record")
+		}
+		h := cr.PCS.ReferenceHasher()
+		_, err = h.Write(buf)
+		if err != nil {
+			return nil, errors.Wrap(err, "[ ContractRequester::Call ] Failed to calculate hash")
+		}
+
+		copy(reqHash[:], h.Sum(nil)[0:insolar.RecordHashSize])
```
ruz: Can we make this a function? A private function within ContractRequester, but function.

ilyapt: ok

PR #2023, opened 2019-07-03
INS-2870 don't use sigkill, it can't be handled properly

aalekseev-ru:
**- What I did**
change signal for proper handling it


aalekseev-ru:
more info - https://github.com/golang/go/issues/9463

PR #2031, opened 2019-07-04
WLT-593: sc result interface to map

pol9kov:
**- What I did**
Edited type of returned result from contract. 
**- How I did it**
 Earlier contracts  returned interface{}, now they returns map[string]interface{}. Also added method migration.createMember that returns map with two objects. And wrote test for this method.
**- How to verify it**
Launch tests
sc result interface to map


api/call.go
```diff
@@ -129,7 +129,7 @@ func processError(err error, extraMsg string, resp *requester.ContractAnswer, in
 func writeResponse(insLog assert.TestingT, response http.ResponseWriter, contractAnswer *requester.ContractAnswer) {
 	res, err := json.MarshalIndent(*contractAnswer, "", "    ")
 	if err != nil {
-		res = []byte(`{"error": "can't marshal ContractAnswer to json'"}`)
+		res = []byte(`{"jsonrpc": "2.0", "id": 1, "error": "can't marshal ContractAnswer to json; error: '` + err.Error() + `'"}`)
```
Hirama: id param should not be hardcoded

pol9kov:
When you explicitly set the map - everything is fine, but when this map is defined as an interface - json marshaling/unmarshaling breaks down. 
For example: it reads this map as map [interface {}] interface {}. The key of map can not be interface so it returns error "json: unsupported type: map [interface {}] interface {}".

PR #2038, opened 2019-07-05
INS-2896: fixed json marshal; added migration.createMember for test fix

pol9kov:
**- What I did**
Fixed error with marshaling map - result from contract
**- How I did it**
Import new dependency - jsoniter. Edited contract for return map (according to Open-API). Wrote test for map return.
**- How to verify it**
Launch test TestMigrationCreateMember
closes #2896

### DIFF
--- a/api/call.go
+++ b/api/call.go
@@ -21,11 +21,13 @@ import (
 	"crypto/sha256"
 	"encoding/base64"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"strings"
 	"time"
 
+	jsoniter "github.com/json-iterator/go"
 	"github.com/stretchr/testify/assert"
 
 	"github.com/insolar/insolar/metrics"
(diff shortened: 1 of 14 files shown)


api/call.go
```diff
@@ -127,9 +129,9 @@ func processError(err error, extraMsg string, resp *requester.ContractAnswer, in
 }
 
 func writeResponse(insLog assert.TestingT, response http.ResponseWriter, contractAnswer *requester.ContractAnswer) {
-	res, err := json.MarshalIndent(*contractAnswer, "", "    ")
+	res, err := jsoniter.MarshalIndent(*contractAnswer, "", "    ")
```
ivanshibitov: What's wrong with standard json lib?

api/call.go
```diff
@@ -127,9 +129,9 @@ func processError(err error, extraMsg string, resp *requester.ContractAnswer, in
...
 	if err != nil {
-		res = []byte(`{"error": "can't marshal ContractAnswer to json'"}`)
+		res = []byte(fmt.Sprintf(`{"jsonrpc": "2.0", "id": %v, "error": {"message": "can't marshal ContractAnswer to json; error: '%v'"}}`, contractAnswer.ID, err.Error()))
```
ivanshibitov: Why do we need write all these fields manually?

logicrunner/builtin/contract/rootdomain/rootdomain.go
```diff
@@ -145,6 +145,16 @@ func (rd *RootDomain) AddNewMemberToMaps(publicKey string, burnAddress string, m
 	return nil
 }
 
+// AddNewMemberToMaps adds new member to PublicKeyMap
```
ivanshibitov: AddNewMemberToMaps -> AddNewMemberToPublicKeyMap
Why not just AddNewMember ?

PR #2039, opened 2019-07-05
INS-2877: fix overtake jetkeeper logic

uhzxl:
**- What I did**
I made a patch that fixes INS-2877.
**- How I did it**
I made a condition that we update only the subsequent pulse or the second pulse after the start of the network. And I fix jet tree actualizing when we handle hot data on heavy node.
**- How to verify it**
Run launchnet + run unit tests.

### DIFF
--- a/insolar/jet/dbstore.go
+++ b/insolar/jet/dbstore.go
@@ -87,7 +87,7 @@ func (s *DBStore) Clone(ctx context.Context, from, to insolar.PulseNumber) error
 	defer s.Unlock()
 
 	tree := s.get(from)
-	newTree := tree.Clone(false)
+	newTree := tree.Clone(true)
 	err := s.set(to, newTree)
 	if err != nil {
 		return errors.Wrapf(err, "failed to clone jet.Tree")
--- a/insolar/jet/dbstore.go
+++ b/insolar/jet/dbstore.go
@@ -82,12 +82,12 @@ func (s *DBStore) Split(ctx context.Context, pulse insolar.PulseNumber, id insol
 	}
 	return left, right, nil
 }
-func (s *DBStore) Clone(ctx context.Context, from, to insolar.PulseNumber) error {
+func (s *DBStore) Clone(ctx context.Context, from, to insolar.PulseNumber, keep bool) error {
 	s.Lock()
 	defer s.Unlock()
 
 	tree := s.get(from)
-	newTree := tree.Clone(true)
+	newTree := tree.Clone(keep)
 	err := s.set(to, newTree)
 	if err != nil {
 		return errors.Wrapf(err, "failed to clone jet.Tree")
--- a/ledger/heavy/handler/handler.go
+++ b/ledger/heavy/handler/handler.go
@@ -55,11 +55,12 @@ type Handler struct {
 	IndexAccessor object.IndexAccessor
 	IndexModifier object.IndexModifier
 
-	DropModifier  drop.Modifier
-	PulseAccessor pulse.Accessor
-	JetModifier   jet.Modifier
-	JetKeeper     replica.JetKeeper
-	Sender        bus.Sender
+	DropModifier    drop.Modifier
+	PulseAccessor   pulse.Accessor
+	PulseCalculator pulse.Calculator
+	JetModifier     jet.Modifier
+	JetKeeper       replica.JetKeeper
+	Sender          bus.Sender
 
 	jetID insolar.JetID
 	dep   *proc.Dependencies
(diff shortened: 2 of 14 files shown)


ledger/heavy/handler/handler.go
```diff
@@ -364,24 +365,32 @@ func (h *Handler) handleHeavyPayload(ctx context.Context, genericMsg insolar.Par
...
 		return &reply.HeavyError{Message: err.Error(), JetID: msg.JetID, PulseNum: msg.PulseNum}, nil
 	}
-	if drop.Split {
-		_, _, err = h.JetModifier.Split(ctx, futurePulse, drop.JetID)
+	pulse := drop.Pulse
 
-	} else {
-		err = h.JetModifier.Update(ctx, futurePulse, false, drop.JetID)
+	// We can't to use PulseCalculator here because split pulse may exceed latest pulse.
+	whenSplit := pulse + 20
```
LinnikD: What is 20 here? If it is 2 next pulses - it is wrong, because pulse length can be changed on pulsar

ledger/heavy/replica/jetkeeper.go
```diff
@@ -169,3 +189,12 @@ func (jk *dbJetKeeper) updateSyncPulse(pn insolar.PulseNumber) error {
 	err := jk.db.Set(syncPulseKey(pn), []byte{})
 	return errors.Wrapf(err, "failed to set up new sync pulse")
 }
+
+func (jk *dbJetKeeper) topSyncPulse() insolar.PulseNumber {
```
LinnikD: This function is hard to understand.
In discussion with @andreyromancev he suggest to store last synced pulse in constant key and update it

ledger/heavy/handler/handler.go
uhzxl: Of course, I know and I wrote it just to make it clear and could be discussed. Any suggestions on how to get around?

ledger/heavy/replica/jetkeeper.go
andreyromancev: +

insolar/jet/dbstore.go
```diff
@@ -87,7 +87,7 @@ func (s *DBStore) Clone(ctx context.Context, from, to insolar.PulseNumber) error
 	defer s.Unlock()
 
 	tree := s.get(from)
-	newTree := tree.Clone(false)
+	newTree := tree.Clone(true)
```
andreyromancev: This is not consistant with In-memory version. If we need to keep cloned tree, let's pass flag as argument for both versions.

ledger/heavy/handler/handler.go
andreyromancev: This is not correct. We should never assume the number of two pulses ahead. We can only assume that NextPulseNumber is correct.

andreyromancev: We need to fetch pulse from drop and take NextPulseNumber from it. It should be the split pulse, not two pulses ahead.

ledger/heavy/handler/handler.go
```diff
@@ -398,3 +407,37 @@ func (h *Handler) handleHeavyPayload(ctx context.Context, genericMsg insolar.Par
...
+	next, err := h.PulseCalculator.Forwards(ctx, from, 1)
+	if err == pulse.ErrNotFound {
+		return
+	}
+	if err != nil {
+		logger.Error(errors.Wrapf(err, "failed to get next pulse number"))
+		return
+	}
+	for to := next.PulseNumber; to <= latest.PulseNumber; {
+		err := h.JetModifier.Clone(ctx, from, to)
```
andreyromancev: This is wrong. We should never clone tree and override existing trees.

ledger/heavy/pulsemanager/pulsemanager.go
```diff
@@ -161,12 +157,19 @@ func (m *PulseManager) setUnderGilSection(ctx context.Context, newPulse insolar.
...
-			err := m.JetModifier.Update(ctx, futurePulse, false, insolar.ZeroJetID)
+			err := m.JetModifier.Update(ctx, newPulse.PulseNumber, true, insolar.ZeroJetID)
 			if err != nil {
 				return errors.Wrapf(err, "failed to update zeroJet")
 			}
 			logger.Infof("[PulseManager] activate zeroJet pulse=%v", futurePulse)
 		}
+
+		futurePulse := newPulse.NextPulseNumber
+		logger.Warnf("CLONE %v -> %v", newPulse.PulseNumber, futurePulse)
```
andreyromancev: Info

ledger/heavy/replica/jetkeeper.go
uhzxl: done

ledger/heavy/handler/handler.go
uhzxl: Seems like jet.Split actually performed on the second pulse after drop.Pulse: https://github.com/insolar/insolar/blob/57ef5bbd8ba11be962a8003adca18cd2d7324686/ledger/light/executor/jet_splitter.go#L116

ledger/heavy/handler/handler.go
uhzxl: If the drop can't be late, then I'll remove the excessive cloning. Is it?

insolar/jet/dbstore.go
uhzxl: Added flag in both versions

ledger/heavy/pulsemanager/pulsemanager.go
uhzxl: cleared

server/internal/heavy/components.go
```diff
@@ -256,9 +256,11 @@ func newComponents(ctx context.Context, cfg configuration.Configuration, genesis
 		h.DropModifier = drops
 		h.PCS = CryptoScheme
 		h.PulseAccessor = Pulses
+		h.PulseCalculator = Pulses
 		h.JetModifier = jets
 		h.JetKeeper = jetKeeper
 		h.Sender = WmBus
+		h.PulseDelta = insolar.PulseNumber(cfg.Pulsar.NumberDelta)
```
egorikas: We can't use pulsar config inside a heavy node. Pulse is a completely standalone entity. You can see, that during working process of the network, Pulsar section of configs are empty for all nodes, except Pulsar

PR #2051, opened 2019-07-08
Ins 2896 struct return

pol9kov:
**- What I did**
Added return struct from smart contracts feature.

**- How I did it**
Added cast all map types to map[string]interface{} in serialization function.

**- How to verify it**
Launch TestCallHelloWorldReturnObj .

struct return from contract


pol9kov:
Лишние коммиты в ветке из мастера

PR #2065, opened 2019-07-09
INS-2932: reason added to requests from ContractService

ilyapt:
**- What I did**
- reason added to requests from ContractService

### DIFF
--- a/api/contract.go
+++ b/api/contract.go
@@ -142,7 +142,7 @@ func (s *ContractService) CallConstructor(r *http.Request, args *CallConstructor
 			Prototype:       protoRef,
 			CallType:        record.CTSaveAsChild,
 			APIRequestID:    utils.TraceID(ctx),
-			Reason:          *insolarApi.MakeReason(pulse.PulseNumber, args.MethodArgs),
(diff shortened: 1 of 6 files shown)


api/contract.go
```diff
@@ -181,13 +193,24 @@ func (s *ContractService) CallMethod(r *http.Request, args *CallMethodArgs, re *
...
+	}
+
 	msg := &message.CallMethod{
 		IncomingRequest: record.IncomingRequest{
 			Caller:       testutils.RandomRef(),
 			Object:       objectRef,
 			Method:       args.Method,
 			Arguments:    args.MethodArgs,
 			APIRequestID: utils.TraceID(ctx),
+			Reason:       *insolar.NewReference(*insolar.NewID(pulse.PulseNumber, hash.Sum(nil))),
```
aalekseev-ru: can we make special method for generate reason? I see copy-paste code, it can produce bug's later, when you change code in one place and doesn't change in another.

aalekseev-ru: And i'm sure, that reason generates not only in this two places

ilyapt: done

PR #2077, opened 2019-07-09
INS-2008: remove parcel from as many places as possible

ruz:


logicrunner/handle_executorresults.go
```diff
@@ -86,8 +86,8 @@ func (p *initializeExecutionState) Proceed(ctx context.Context) error {
 	// prepare Queue
 	if p.msg.Queue != nil {
 		for _, qe := range p.msg.Queue {
-			ctxToSent := qe.Parcel.Context(context.Background())
-			transcript := NewTranscript(ctxToSent, qe.Parcel, qe.Request)
+			ctxToSent := context.TODO() //FIXME: !!!!
```
aalekseev-ru: why not context.Background() as it was before?

ruz: created a task, INS-2944

PR #2094, opened 2019-07-11
INS-2944: restore ServiceData when we pass requests between virtuals

ruz:
* cover ExecutorResults containing requests for execution

* cover AdditionalCallFromPreviousExecutor

* correctly switch to transcript.Context in the broker when executing

### DIFF
--- a/logicrunner/executionbroker.go
+++ b/logicrunner/executionbroker.go
@@ -344,6 +344,12 @@ func (q *ExecutionBroker) finishTask(ctx context.Context, transcript *Transcript
 }
 
 func (q *ExecutionBroker) processTranscript(ctx context.Context, transcript *Transcript) bool {
+	if transcript.Context != nil {
+		ctx = transcript.Context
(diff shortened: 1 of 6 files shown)


logicrunner/executionbroker.go
```diff
@@ -343,6 +343,12 @@ func (q *ExecutionBroker) finishTask(ctx context.Context, transcript *Transcript
 }
 
 func (q *ExecutionBroker) processTranscript(ctx context.Context, transcript *Transcript) bool {
+	if transcript.Context != nil {
+		ctx = transcript.Context
+	} else {
+		inslogger.FromContext(ctx).Error("context in transcript is nil")
+	}
+
```
afiskon: Any chance we could fill ctx with context.Default() in tests instead?

ruz: we decided to keep this check, howere in tests I made sure we don't have this error

PR #2131, opened 2019-07-15
Ins 2972 wait ok sender

megge-dream:
**- What I did**
Add `WaitOKSender` struct, which allows to wait for one reply, and if its not OK, to log error.
`NewWaitOKWithRetrySender` func creates `WaitOKSender` instance which retries `CodeFlowCanceled` errors.

**- How to verify it**
Run tests.

### DIFF
--- a/insolar/bus/waiter.go
+++ b/insolar/bus/waiter.go
@@ -49,8 +49,8 @@ func NewWaitOKSender(sender Sender) *WaitOKSender {
 	return c
 }
 
-// SendRole sends message to specified role, using provided Sender.SendRole. It waiting for reply.OK and
-// close replies channel after getting it. If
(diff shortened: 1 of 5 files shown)


insolar/bus/waiter.go
```diff
@@ -0,0 +1,116 @@
...
+		return
+	}
+
+	switch p := replyPayload.(type) {
+	case *payload.Error:
+		logger.Errorf("expected OK, got error: %s, code - %d", p.Text, p.Code)
+	default:
+		logger.Errorf("got unexpected reply: %#v", p)
+	}
+	return
```
golangcibot: S1023: redundant `return` statement (from `gosimple`)

PR #2149, opened 2019-07-17
Ins 2877 jet keeper awear of pulse order

ivanshibitov:
**- What I did**

**- How I did it**

**- How to verify it**


ledger/heavy/executor/jetkeeper.go
```diff
@@ -84,34 +129,38 @@ func (jk *dbJetKeeper) add(pulse insolar.PulseNumber, id insolar.JetID) error {
...
 		}
+		return false
 	}
 
-	err := jk.updateSyncPulse(pulse)
-	if err != nil {
-		return errors.Wrapf(err, "failed to update consistent pulse")
+	actualJetsSet := toSet(actualJets)
+
+	for k, _ := range toSet(expectedJets) {
```
golangcibot: should omit 2nd value from range; this loop is equivalent to `for k := range ...` (from `golint`)

PR #2160, opened 2019-07-17
INS-2992: system errors

afiskon:
Current limitations:
* The platform doesn't has a proper support of constructors that return errors
* Resursive calls are considered a system error, not a contract error
* Porotype mismatch is always considered a system error
* External call from immutable method is considered a system error

These issues will be fixed separately.


logicrunner/common/systemerror.go
```diff
@@ -0,0 +1,34 @@
...
+const glsSystemErrorKey = "systemError"
+
+type SystemError interface {
+	GetSystemError() error
+	SetSystemError(err error)
+}
+
+type systemError struct{}
+
+func NewSystemError() *systemError {
```
golangcibot: exported func NewSystemError returns unexported type *common.systemError, which can be annoying to use (from `golint`)

afiskon: tnx fixed

PR #2172, opened 2019-07-18
INS-2971: Copy LR messages in protobuf format

aalekseev-ru:
**- What I did**
add protobuff for all LR handlers messages
**- How I did it**
add them into payload.proto, regen payload.pb.go, add them into payload.go, regen strings. Also move pending state from message package to insolar package
**- How to verify it**
run tests


insolar/payload/payload.proto
```diff
@@ -9,11 +9,11 @@ import "github.com/insolar/insolar/insolar/record/record.proto";
 message Meta {
     uint32 Polymorph = 16;
 
-    bytes Payload    = 20;
```
bigbes: return whitespaces, pls

aalekseev-ru: @andreyromancev  is there some rules for formating this file?

PR #2177, opened 2019-07-18
WLT-727: edited getBalance method

pol9kov:
**- What I did**
Added deposits to getBalance method like in openAPI
**- How I did it**
Changed getBalance method; changed deposit marshal to map; add struc GetBalanceResponse. 
**- How to verify it**
We can test getBalance deposits only in migration tests when migration will be done

GetBalance method update

### DIFF
--- a/logicrunner/builtin/contract/deposit/deposit.go
+++ b/logicrunner/builtin/contract/deposit/deposit.go
@@ -81,7 +81,8 @@ func (d *Deposit) MapMarshal() (map[string]string, error) {
 		"holdReleaseDate": d.HoldReleaseDate.String(),
 		"amount":          d.Amount,
 		"bonus":           d.Bonus,
-		"txId":            d.TxHash,
+		"ethTxHash":       d.TxHash,
+		"status":          string(d.Status),
 	}, nil
 }
 
--- a/logicrunner/builtin/contract/deposit/deposit.go
+++ b/logicrunner/builtin/contract/deposit/deposit.go
@@ -75,7 +75,7 @@ func New(migrationDaemonConfirms map[insolar.Reference]bool, txHash string, amou
 }
 
 // MapMarshal gets deposit information.
-func (d *Deposit) Itself() (Deposit, error) {
+func (d *Deposit) Itself() (interface{}, error) {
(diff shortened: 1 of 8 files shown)


logicrunner/builtin/contract/member/member.go
```diff
@@ -458,50 +465,67 @@ func (m *Member) migration(txHash string, burnAddress string, amount big.Int, un
...
-// FindDeposit finds deposits for this member with this transaction hash.
-func (m *Member) FindDeposit(txHash string, inputAmountStr string) (bool, deposit.Deposit, error) {
+// GetDeposits get all deposits for this member
+func (m *Member) GetDeposits() ([]map[string]string, error) {
+	var result []map[string]string
+	for _, dRef := range m.Deposits {
+
+		d := deposit.GetObject(dRef)
+
+		depositInfo, err := d.MapMarshal()
```
LinnikD: Its better to use smth like DepositResoinse object with tags for json marshaling in GetBalanceResponse

logicrunner/builtin/contract/deposit/deposit.go
```diff
@@ -75,14 +75,8 @@ func New(migrationDaemonConfirms map[insolar.Reference]bool, txHash string, amou
...
 // MapMarshal gets deposit information.
-func (d *Deposit) MapMarshal() (map[string]string, error) {
-	return map[string]string{
-		"timestamp":       d.Timestamp.String(),
-		"holdReleaseDate": d.HoldReleaseDate.String(),
-		"amount":          d.Amount,
-		"bonus":           d.Bonus,
-		"txId":            d.TxHash,
-	}, nil
+func (d *Deposit) Itself() (Deposit, error) {
```
LinnikD: Why not `Get`?

pol9kov: Because get what?

pol9kov: Mby GetItself? But we can avoid get I think.

pol9kov: But I thing Itself() is better

PR #2181, opened 2019-07-19
Wlt 503 replicator draft

andreyromancev:
**- What I did**

**- How I did it**

**- How to verify it**


configuration/ledger.go
```diff
@@ -52,6 +52,24 @@ type Exporter struct {
 	ExportLag uint32
 }
 
+// Replica holds configuration for Replicator.
+type Replica struct {
+	// Role defines that should do replicator (subscribe on replica parent or send notifications to replica targets).
+	Role string
```
egorikas: It's strange for me, that we have a part of observer functionality inside a heavy's node. Because observer is a standalone node

insolar/noderole.go
```diff
@@ -24,13 +24,17 @@ const (
 	StaticRoleVirtual
 	StaticRoleHeavyMaterial
 	StaticRoleLightMaterial
+	StaticRoleHeavyReplica
```
egorikas: The same situation. We don't have a new role inside a system. Obeserver is a standalone node

insolar/pulse/db.go
```diff
@@ -89,6 +89,7 @@ func (s *DB) Append(ctx context.Context, pulse insolar.Pulse) error {
 		if err != nil {
 			return err
 		}
+		oldHead.Pulse.NextPulseNumber = pulse.PulseNumber
```
egorikas: We mustn't modify `Pulse` struct. We have a sign inside of it

```go
// Pulse is base data structure for a pulse.
type Pulse struct {
	PulseNumber     PulseNumber
	PrevPulseNumber PulseNumber
	NextPulseNumber PulseNumber

	PulseTimestamp   int64
	EpochPulseNumber int
	OriginID         [OriginIDSize]byte

	Entropy Entropy
	Signs   map[string]PulseSenderConfirmation
}

```

internal/ledger/store/db.go
```diff
@@ -40,7 +40,7 @@ type Iterator interface {
 	// Value returns value itself (ex: record, drop, blob, etc).
 	// Warning: Value is only valid as long as item is valid (until iterator.Next() called), or transaction is valid.
 	// If you need to use it outside its validity, please copy the value.
-	Value() ([]byte, error)
```
egorikas: We mustn't remove the result error here. It's a ok situation in go to return err, when iterator is in the an end

internal/ledger/store/badger.go
```diff
@@ -247,6 +247,6 @@ func (bi *badgerIterator) Key() []byte {
 	return bi.prevKey
 }
 
-func (bi *badgerIterator) Value() ([]byte, error) {
-	return bi.prevValue, nil
+func (bi *badgerIterator) Value() []byte {
```
egorikas: See comment about Iterator interface below

PR #2184, opened 2019-07-19
Ins 2997 get jet

aMFeriuSS:
**- What I did**
new GetJet handler and procedure are used by JetFetcher + jetfetcher replacing


ledger/light/executor/jet_fetcher.go
```diff
@@ -228,50 +229,67 @@ func (tu *fetcher) fetch(
...
+					once.Do(func() {
+						jID := concrete.JetID
+						jetID := insolar.ID(jID)
+						ch <- fetchResult{&jetID, nil}
+						close(ch)
+					})
+
+					replies[i] = concrete.JetID
+				case *payload.Error:
+					return
```
andreyromancev: Let's log error

aMFeriuSS: Right! I forgot this. But now it's done

PR #2199, opened 2019-07-22
INS-3043: proper node leaving with disabled consensus

Vany:


ivanshibitov:
What does it do? =)
What does it fix?

PR #2209, opened 2019-07-23
INS-3057: bench fix

pol9kov:
**- What I did**
Fixed bench error when repetition >= 1
**- How I did it**
Changed calculation of total balance after transfers. Now multi repetition with fee
**- How to verify it**
launch ./scripts/bench.sh -r=2

bench fix

### DIFF
--- a/cmd/benchmark/main.go
+++ b/cmd/benchmark/main.go
@@ -401,7 +401,7 @@ func main() {
 		totalBalanceAfterWithFee := big.NewInt(0)
 		for nretries := 0; nretries < 3; nretries++ {
 			totalBalanceAfter, _ = getTotalBalance(insSDK, members)
-			totalBalanceAfterWithFee = new(big.Int).Add(totalBalanceAfter, big.NewInt(calcFee(transferAmount)*int64(repetitions)))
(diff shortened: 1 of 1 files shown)


cmd/benchmark/main.go
```diff
@@ -401,7 +401,7 @@ func main() {
 		totalBalanceAfterWithFee := big.NewInt(0)
 		for nretries := 0; nretries < 3; nretries++ {
 			totalBalanceAfter, _ = getTotalBalance(insSDK, members)
-			totalBalanceAfterWithFee = new(big.Int).Add(totalBalanceAfter, big.NewInt(calcFee(transferAmount)))
+			totalBalanceAfterWithFee = new(big.Int).Add(totalBalanceAfter, big.NewInt(calcFee(transferAmount)*int64(repetitions)))
```
megge-dream: Repetitions is not equal to count of transfer operations. Repetitions - count of transfer for one user (as it written in flag description). Count of transfer operations will be `repetitions*concurrent`.

pol9kov: Thank you! fixed

PR #2226, opened 2019-07-25
INS-3033: add correct check for uploadCode hackish requests

bigbes:


insolar/record/record.go
```diff
@@ -204,6 +211,10 @@ func (r *OutgoingRequest) IsDetached() bool {
 	return isDetached(r.ReturnMode)
 }
 
+func (m *OutgoingRequest) IsTemporaryUploadCode() bool {
```
golangcibot: receiver name m should be consistent with previous receiver name r for OutgoingRequest (from `golint`)

PR #2251, opened 2019-07-26
either Caller or APINode should be set, not both

ruz:


contractrequester/contractrequester.go
```diff
@@ -131,6 +131,17 @@ func (cr *ContractRequester) calcRequestHash(request record.IncomingRequest) ([i
 	return hash, nil
 }
 
+func (cr *ContractRequester) checkCall(ctx context.Context, msg *message.CallMethod) error {
```
golangcibot: `(*ContractRequester).checkCall` - `ctx` is unused (from `unparam`)

PR #2259, opened 2019-07-26
INS-3051: Remove extra fields from index.

andreyromancev:


ledger/object/lifeline.go
```diff
@@ -54,33 +54,10 @@ func CloneLifeline(idx record.Lifeline) record.Lifeline {
...
-	} else {
-		idx.Delegates = []record.LifelineDelegate{}
-	}
-
 	if idx.EarliestOpenRequest != nil {
 		tmp := *idx.EarliestOpenRequest
 		idx.EarliestOpenRequest = &tmp
 	}
 
-	if idx.PendingPointer != nil {
```
ruz: forgot to clone LatestRequest

internal/ledger/artifact/artifact.go
```diff
@@ -160,35 +159,17 @@ func (m *Scope) activateObject(
 	parent insolar.Reference,
 	memory []byte,
 ) error {
-	parentIdx, err := m.IndexAccessor.ForID(ctx, m.PulseNumber, *parent.Record())
```
ruz: we kept parent check in artifacts.client... to make sure parent reference is valid and not bogus

andreyromancev: Genesis doesn't have any checks. It just writes data into storage from config.

ruz: this if below makes sure that the parent exists. what is wrong with keeping these four lines here?

PR #2290, opened 2019-07-31
INS-3122: Root domain ID in refs

ita-sammann:
**- What I did**
* Add root domain ID to references.
* Preserve domain when unmarshalling reference from a string.


ita-sammann:
Will be done later

PR #2292, opened 2019-07-31
WLT-775: added shard; edited MA and PK to shards

pol9kov:
**- What I did**
Added shards. Changed all functionality to shards.

**- How I did it**
Added shard contract; create 1000 shards for migration addresses and 1000 shards for public keys on bootstrap; set references to arrays of shards in rootdomain; changed functionality to shard-usage

**- How to verify it**
Launch tests

added shards


logicrunner/builtin/contract/shard/shard.go
```diff
@@ -0,0 +1,54 @@
...
+// Shard - basic shard contract.
+type Shard struct {
+	foundation.BaseContract
+	Map foundation.StableMap
+}
+
+// New creates new member.
+func New() (*Shard, error) {
+	return &Shard{
+		Map: map[string]string{},
```
LinnikD: You should use StableMap instead

logicrunner/builtin/contract/rootdomain/rootdomain.go
```diff
@@ -85,31 +87,43 @@ func (rd *RootDomain) GetBurnAddress() (string, error) {
...
+	return ref, nil
 }
 
-// GetMemberByBurnAddress gets member reference by burn address.
-func (rd RootDomain) GetMemberByBurnAddress(burnAddress string) (insolar.Reference, error) {
-	result, ok := rd.BurnAddressMap[trimBurnAddress(burnAddress)]
-	if !ok {
-		return insolar.Reference{}, fmt.Errorf("member for this migration address does not exist")
+// GetMemberByMigrationAddress gets member reference by burn address.
+func (rd RootDomain) GetMemberByMigrationAddress(migrationAddress string) (*insolar.Reference, error) {
```
LinnikD: why by pointer insolar.Reference?
make (rd RootDomain) at pointer

LinnikD: why do we use root domain fo this case?

logicrunner/builtin/contract/rootdomain/rootdomain.go
```diff
@@ -159,30 +173,45 @@ func (rd *RootDomain) AddBurnAddress(burnAddress string) error {
...
 
-// AddNewMemberToMaps adds new member to PublicKeyMap and BurnAddressMap.
-func (rd *RootDomain) AddNewMemberToMaps(publicKey string, burnAddress string, memberRef insolar.Reference) error {
+// AddNewMemberToMaps adds new member to PublicKeyMap and MigrationAddressMap.
+func (rd *RootDomain) AddNewMemberToMaps(publicKey string, migrationAddress string, memberRef insolar.Reference) error {
 	trimmedPublicKey := trimPublicKey(publicKey)
-	if _, ok := rd.PublicKeyMap[trimmedPublicKey]; ok {
-		return fmt.Errorf("member for this publicKey already exist")
+	i := foundation.GetShardIndex(trimmedPublicKey, insolar.GenesisAmountPublicKeyShards)
+	if i >= len(rd.PublicKeyShards) {
```
LinnikD: why do we need this check?

logicrunner/builtin/contract/rootdomain/rootdomain.go
pol9kov: It returns pointer because it is more beautiful to return nil except insolar.Reference{}. 
why make rd pointer? this method doesn't change anything

logicrunner/builtin/contract/rootdomain/rootdomain.go
pol9kov: Because maybe sometime we might break GetShardIndex and it is better to check in contract that all is correct to avoid panic

logicrunner/builtin/contract/shard/shard.go
pol9kov: There is no difference but I can change it to *make(foundation.StableMap)* if you want

PR #2298, opened 2019-07-31
Ins 3099 nonretryable errors

sexyhamster:
**- What I did**
added nonretryable system errors

**- How I did it**
added nonretryable system errors

**- How to verify it**
run tests for "index not found"


utils/errors/errors.go
```diff
@@ -0,0 +1,44 @@
...
+
+package errors
+
+// nonRepeatableError is error that will not be retried
+type nonRetryableError struct {
+	reason string
+}
+
+// NewNonRetryable creates new nonRepeatableError that will not be retried
+func NewNonRetryable(reason string) *nonRetryableError {
```
golangcibot: exported func NewNonRetryable returns unexported type *errors.nonRetryableError, which can be annoying to use (from `golint`)

sexyhamster:
removed logs

functest/get_balance_test.go
```diff
@@ -37,5 +38,5 @@ func TestGetBalance(t *testing.T) {
 func TestGetBalanceWrongRef(t *testing.T) {
 	_, err := getBalance(&root, testutils.RandomRef().String())
 	require.NotNil(t, err)
-	require.Contains(t, err.Error(), "index not found")
+	require.True(t, errors.IsNonRetryable(err), "index not found")
```
megge-dream: The last param of `require.True` is the additional message to tests trace in case of fail. Why is it says `index not found`?

functest/send_money_test.go
```diff
@@ -70,7 +72,7 @@ func TestTransferMoneyFromNotExist(t *testing.T) {
 
 	_, err := signedRequest(firstMember, "member.transfer", map[string]interface{}{"amount": amount, "toMemberReference": secondMember.ref})
 	require.NotNil(t, err)
-	require.Contains(t, err.Error(), "index not found")
+	require.True(t, errors.IsNonRetryable(err), "index not found")
```
megge-dream: The last param of `require.True` is the additional message to tests trace in case of fail. Why is it says `index not found`?

functest/send_money_test.go
```diff
@@ -84,7 +86,7 @@ func TestTransferMoneyToNotExist(t *testing.T) {
 
 	_, err := signedRequest(firstMember, "member.transfer", map[string]interface{}{"amount": amount, "toMemberReference": testutils.RandomRef().String()})
 	require.NotNil(t, err)
-	require.Contains(t, err.Error(), "index not found")
+	require.True(t, errors.IsNonRetryable(err), "index not found")
```
megge-dream: The last param of `require.True` is the additional message to tests trace in case of fail. Why is it says `index not found`?

functest/get_balance_test.go
megge-dream: We send API request here, so you don't get direct instance of `Retryable` error, you get some json with `error` field inside, and convert it to basic `error` interface for working with it like its object.

functest/send_money_test.go
megge-dream: We send API request here, so you don't get direct instance of `Retryable` error, you get some json with `error` field inside, and convert it to basic `error` interface for working with it like its object.

functest/send_money_test.go
megge-dream: We send API request here, so you don't get direct instance of `Retryable` error, you get some json with `error` field inside, and convert it to basic `error` interface for working with it like its object.

functest/get_balance_test.go
sexyhamster: reverted this part

sexyhamster:
we have better version

PR #2304, opened 2019-08-01
Wlt 776 merge tariff

Hirama:
**- What I did**
Removed tariff, now cost centre provide fee rate and calculations are moved to a wallet contract


pol9kov:
Well done

PR #2307, opened 2019-08-02
Ins 3081 new logicrunner tests

Tsovak:
**- What I did**
new logicrunner tests
**- How I did it**

**- How to verify it**


functest/test-data/simple_contract_with_sleep.txt
```diff
@@ -0,0 +1,47 @@
+package main
```
LinnikD: why not to move all contracts to separate files?

ruz: what? cuz code of contract crucial to the test and it's way harder to understand test when code is further away.

ruz: Please don't do this.

ruz: keep style of the tests intact and follow it. if you want to introduce functest/test-data/ directory then do it in a separate branch and for all tests.

PR #2318, opened 2019-08-05
NOISSUE: add ExecutionArchive tests

bigbes:


insolar/gen/string.go
```diff
@@ -33,3 +36,11 @@ func StringFromBytes(symbols []byte, maxcount int) string {
 	f.NilChance(0).Fuzz(&bstr)
 	return string(bstr)
 }
+
+func APIRequestID() string {
```
ruz: `APIRequestID: utils.RandTraceID(),`

We already have function. Why do you need new?

bigbes: gen is package that superseed old utils/testutils functions, as i've understand

bigbes: BTW, traceid != apirequestid, so its not fully "transparant", why traceid used for apirequestid field.

ruz: Not really. Gen is to generate with fuzz and shit like this for tests. You shouldn't use `gen` in production code. However, in this particular case of generating `APIRequestID`, we generate the field in production, so we use an utility function for that. I find it ok to use this utility function in tests.

bigbes: Still, i don't think using `utils.RandTraceID()` as `APIRequestID` is appropriate. I can use it now, but it's still inconsistency.

ruz: May be you're right and either works.

PR #2331, opened 2019-08-05
INS-3059: serialize constructor errors' as result of constructor

ruz:
make constructor and methods even more similar, now both return two
things: NewState and Result.

State is serialized memory

Result is serilized returns of the function including error.

With constructors it can be only error, signatur of construcor a bit
limitted to state+error ATM


logicrunner/builtin/proxyhelper.go
```diff
@@ -73,22 +71,26 @@ func (h *ProxyHelper) RouteCall(ref insolar.Reference, wait bool, immutable bool
...
 		h.SetSystemError(err)
 		return nil, err
 	}
+
 	return res.Result, nil
 }
 
-func (h *ProxyHelper) SaveAsChild(parentRef, classRef insolar.Reference, constructorName string,
-	argsSerialized []byte) (objRef insolar.Reference, err error) {
+func (h *ProxyHelper) SaveAsChild(
```
sexyhamster: i think it worth a comment what is returned, because now it's not obvious

logicrunner/requestsexecutor_test.go
```diff
@@ -90,7 +98,7 @@ func TestRequestsExecutor_Execute(t *testing.T) {
 	ctx := inslogger.TestContext(t)
 	mc := minimock.NewController(t)
 	defer mc.Finish()
-	defer mc.Wait(time.Minute)
+	defer mc.Wait(time.Second)
```
bigbes: sure?

logicrunner/goplugin/rpctypes/rpctypes.go
```diff
@@ -51,8 +51,8 @@ type DownCallConstructorReq struct {
 
 // DownCallConstructorResp is response from CallConstructor RPC in the runner
 type DownCallConstructorResp struct {
-	Ret              insolar.Arguments
-	ConstructorError string // gob can't serialize `error` thus we are using string here
+	Data []byte
+	Ret  insolar.Arguments
```
bigbes: Maybe rename this `Ret` to `Result`, for instance?
Sure it should have type `insolar.Arguments`?

ruz: put on my cleanup list

PR #2348, opened 2019-08-06
INS-3061: OnPulse function call order test

bigbes:


logicrunner/writecontroller/writecontroller.go
```diff
@@ -42,34 +41,41 @@ type Accessor interface {
...
+type writeController struct {
 	lock    sync.RWMutex
 	current insolar.PulseNumber
 	closed  bool
 
 	wg *sync.WaitGroup
 }
 
-func NewWriteController() *WriteController {
+func NewWriteController() *writeController {
```
golangcibot: exported func NewWriteController returns unexported type *writecontroller.writeController, which can be annoying to use (from `golint`)

insolar/flow/dispatcher/dispatcher.go
```diff
@@ -46,13 +46,15 @@ type Dispatcher struct {
 	PulseAccessor insPulse.Accessor
 }
 
-func NewDispatcher(present flow.MakeHandle, future flow.MakeHandle, past flow.MakeHandle) *Dispatcher {
+func NewDispatcher(pulseAccessor insPulse.Accessor, present flow.MakeHandle, future flow.MakeHandle, past flow.MakeHandle) *Dispatcher {
```
sexyhamster: if pulseAccessor is not required, maybe don't put it in constructor? most of the calls in this PR sends nil

maybe put it to setter

but I'm not sure how p, err := d.PulseAccessor.Latest(ctx) works at all

bigbes: discused, fixed

PR #2354, opened 2019-08-07
refactor return from handle calls

ruz:


ruz:
unit tests were failing, had fix them

PR #2356, opened 2019-08-07
WLT-794: added saga in transfer

pol9kov:
**- What I did**
added saga in transfer
**- How I did it**
added saga in transfer
**- How to verify it**
launch test

added saga in transfer


logicrunner/builtin/contract/wallet/wallet.go
```diff
@@ -109,21 +109,16 @@ func (w *Wallet) Transfer(rootDomainRef insolar.Reference, amountStr string, toM
 		return member.TransferResponse{Fee: feeStr}, nil
```
bigbes: Can you replace ok/error path? 
```if err != nil {
   return err
}
return ok```

File: logicrunner/builtin/contract/wallet/wallet.go [hunk 2]
```diff
@@ -147,8 +142,8 @@ func (w *Wallet) Accept(amountStr string) (err error) {
 	return nil
 }
 
-// RollBack rolls back transfer to balance.
-func (w *Wallet) RollBack(amountStr string) (err error) {
+// Rollback rolls back transfer to balance.
```
LinnikD: Rollback is unused now? Do we need it?

LinnikD: It ok, need it for saga calls

logicrunner/builtin/contract/wallet/wallet.go
pol9kov: Done

PR #2371, opened 2019-08-08
Wlt 801 deposit migration return parameter

pavlushev:
What I did:
- added return parameter to deposit.migration


cmd/migrator/util.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func check(msg string, err error) {
+	if err != nil {
+		fmt.Println(msg, err)
+		os.Exit(1)
+	}
+}
+
+func execute(url string, keys memberKeys, request requester.Request) (string, error) {
```
golangcibot: execute - result 1 (error) is always nil (from `unparam`)

cmd/simplerequester/util.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func check(msg string, err error) {
+	if err != nil {
+		fmt.Println(msg, err)
+		os.Exit(1)
+	}
+}
+
+func execute(url string, keys memberKeys, request requester.Request) (string, error) {
```
golangcibot: execute - result 1 (error) is always nil (from `unparam`)

PR #2387, opened 2019-08-08
NOISSUE: split queue processor into two parts

bigbes:


logicrunner/executionbroker.go
```diff
@@ -687,13 +691,6 @@ func (q *ExecutionBroker) AddFreshRequest(
 	q.stateLock.Lock()
 	defer q.stateLock.Unlock()
 
-	select {
-	case <-ctx.Done():
-		inslogger.FromContext(ctx).Debug("pulse changed, skipping")
-		return
-	default:
-	}
-
```
afiskon: I don't quite understand these changes. What are they for?

afiskon:
I'm OK with this patch but better someone else take a look as well.

logicrunner/executionbroker.go
```diff
@@ -722,13 +719,6 @@ func (q *ExecutionBroker) AddAdditionalRequestFromPrevExecutor(
 	q.stateLock.Lock()
 	defer q.stateLock.Unlock()
 
-	select {
-	case <-ctx.Done():
```
ruz: this is not part of this pull request

logicrunner/executionbroker.go
```diff
@@ -687,13 +691,6 @@ func (q *ExecutionBroker) AddFreshRequest(
...
```
ruz: this is not part of this pull request

logicrunner/executionbroker_test.go
```diff
@@ -673,8 +673,6 @@ func TestExecutionBroker_OnPulse(t *testing.T) {
 				objectRef := gen.Reference()
 				broker := NewExecutionBroker(objectRef, nil, nil, nil, nil, nil, nil, nil, nil)
 				broker.pending = insolar.InPending
-				broker.requestsFetcher = NewRequestsFetcherMock(t).
-					FetchPendingsMock.Return()
```
ruz: I can't believe this is correct, especially considering comment to this test. this test was assuring that we start requestsFetcher and you delete exactly this check.

PR #2400, opened 2019-08-09
NOISSUE: update waiter in functests

megge-dream:
**- What I did**
Update `waitForFunction` in tests:
- don't close channel for result
- use `wrapper` func, not goroutine
- timeout on `time.After`, not timer

**- How to verify it**
Run tests


sexyhamster:
fixed this while was working on my PR

PR #2413, opened 2019-08-09
INS-3087: heavy node should raise fatal error on failed replication

nordicdyno:
**- What I did**

heavy node should fail for data consistency reason
(consistency should be achieved after node reload)

### DIFF
--- a/ledger/heavy/proc/replication.go
+++ b/ledger/heavy/proc/replication.go
@@ -87,7 +87,24 @@ func (p *Replication) Proceed(ctx context.Context) error {
 		return fmt.Errorf("unexpected payload %T", pl)
 	}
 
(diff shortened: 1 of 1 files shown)


ledger/heavy/proc/replication.go
```diff
@@ -136,7 +148,7 @@ func storeDrop(
 	}
 	err = drops.Set(ctx, *d)
 	if err != nil {
-		return nil, errors.Wrapf(err, "heavyserver: drop storing failed")
```
ivanshibitov: Why do you remove wrapping?

nordicdyno: it was useless, we already adding context in `store` method

PR #2414, opened 2019-08-09
INS-3122: remove logicrunner.Ref

ita-sammann:
**- What I did**
Removed logicrunner.Ref

### DIFF
--- a/logicrunner/logicrunner.go
+++ b/logicrunner/logicrunner.go
@@ -22,14 +22,11 @@ import (
 	"strconv"
 	"sync"
 
-	"github.com/insolar/insolar/network"
-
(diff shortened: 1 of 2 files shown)


logicrunner/logicrunner.go
```diff
@@ -19,10 +19,11 @@ package logicrunner
 
 import (
 	"context"
-	"github.com/insolar/insolar/network"
 	"strconv"
 	"sync"
 
+	"github.com/insolar/insolar/network"
```
sexyhamster: plz group insolar imports

ita-sammann: dnoe

PR #2421, opened 2019-08-12
INS-3016: loop detected as logic error

ruz:


logicrunner/handle_calls.go
```diff
@@ -189,7 +189,18 @@ func (h *HandleCall) handleActual(
 	}
 
 	if h.checkExecutionLoop(ctx, *requestRef, request) {
-		return nil, errors.New("loop detected")
+		proc := &RecordErrorResult{
```
sexyhamster: Could you please move this to helper like 
HandleLogicError(ctx, error, flow, lr.ArtifactManager, *requestRef, *objRef)
so that it would be easier to apply the same pattern for all other errors?

ruz: I don't think it's applicable in other places. I thought about this and didn't find any cases that can benefit from code re-usage.

ruz:
rebase to fix conflicts

PR #2418, opened 2019-08-09
Noissue logicrunner refactoring

sexyhamster:
**- What I did**
logicrunner v2

**- How I did it**
with my hands tied behind my head

**- How to verify it**
run tests

**- Description for the changelog**

PR #2427, opened 2019-08-12
INS-3185: remove weNext optimization from OnPulse

bigbes:


bigbes:
fixed conflicts after https://github.com/insolar/insolar/pull/2418

PR #2436, opened 2019-08-12
sync.Map in idLocker, it's more suitable

ruz:
what is different from old code besides sync.Map:
* old code was putting operations on content of the entry under map lock
  rather than entry lock
* new code releases map ASAP
* if we work on many different objects then sync.Map should be better
* entry lock should be just Mutex, not RW, we were never using RLock


nordicdyno:
probably race here, and we don't need it right now

PR #2445, opened 2019-08-13
INS-3163: fix goroutines leak in light integration tests

aMFeriuSS:
**- What I did**
Fix goroutines leak in light integration tests: 

**- How I did it**
- Publisher.Close()
- Subscriber.Close() 
into Server.Stop()

**- How to verify it**
- Add some code for leak detection into `Test_BasicOperations` after `t.Parallel()`
```
d := leaktest.Check(t)
defer d()
```
- run this `CGO_ENABLED=1 go test -timeout 0 ./ledger/light/integration -run 'Test_BasicOperations' -race -failfast -count 300`
- wait for test success

### DIFF
--- a/ledger/light/integration/server_test.go
+++ b/ledger/light/integration/server_test.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"crypto"
 	"fmt"
-	"github.com/insolar/insolar/network"
 	"math"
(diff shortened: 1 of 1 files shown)


aMFeriuSS:
Close because there are failed test here. Need to check and fix.

PR #2457, opened 2019-08-13
[Draft] add p256k curve to insolar utility

pavlushev:
What I did:
- added p256k curve to insolar utility
- added  P256K to keyProcessor interface


cmd/insolar/main.go
```diff
@@ -242,32 +253,52 @@ func generateKeysPair() {
...
+	return nil
+}
+
+func getUserRef(sendURL string, rootMember, migrationAdmin bool) string {
+	if rootMember || migrationAdmin {
+		info, err := requester.Info(sendURL)
+		check("[ sendRequest ]", err)
+		if rootMember {
+			return info.RootMember
+		} else {
```
golangcibot: `if` block ends with a `return` statement, so drop this `else` and outdent its block (from `golint`)

PR #2468, opened 2019-08-14
INS-3236: partial release realization

pol9kov:
**- What I did**
Add partial release
**- How I did it**
Compare amount of transfer with available amount
**- How to verify it**
TestDepositTransferToken

### DIFF
--- a/functest/deposit_transfer_test.go
+++ b/functest/deposit_transfer_test.go
@@ -34,7 +34,7 @@ func TestDepositTransferToken(t *testing.T) {
 	member := fullMigration(t, "Eth_TxHash_test")
 
 	firstBalance := getBalanceNoErr(t, member, member.ref)
-	secondBalance := new(big.Int).Add(firstBalance, big.NewInt(100))
+	secondBalance := new(big.Int).Add(firstBalance, big.NewInt(1000))
 
 	var err error
 	for i := 0; i <= 20; i++ {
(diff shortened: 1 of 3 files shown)


functest/deposit_transfer_test.go
```diff
@@ -33,15 +34,24 @@ func TestDepositTransferToken(t *testing.T) {
...
-		_, _, err = makeSignedRequest(member, "deposit.transfer", map[string]interface{}{"amount": "100", "ethTxHash": "Eth_TxHash_test"})
+		_, _, err = makeSignedRequest(member, "deposit.transfer", map[string]interface{}{"amount": "1000", "ethTxHash": "Eth_TxHash_test"})
+		require.Error(t, err)
+		if !strings.Contains(err.Error(), "hold period didn't end") {
+			break
+		}
+	}
+	require.Contains(t, err.Error(), "not enough unholded balance for transfer")
+
+	for i := 0; i <= 20; i++ {
```
megge-dream: Why 20 tries? Why not one sleep for 10 sec, and than one request?

pol9kov: Because maybe pulse will change earlier then after 10 seconds

megge-dream: you wait for pulse change in prev request, so next pulse (for next request) will be in 10 second

PR #2470, opened 2019-08-14
INS-3171: fixed deposit transfer

pol9kov:
**- What I did**
fixed deposit transfer
**- How I did it**
Add waiting for saga received 
**- How to verify it**
Launch test TestDepositTransferToken with count 100

### DIFF
--- a/functest/deposit_transfer_test.go
+++ b/functest/deposit_transfer_test.go
@@ -48,17 +48,7 @@ func TestDepositTransferToken(t *testing.T) {
 	}
 	require.NoError(t, err)
 
-	var finalBalance *big.Int
-	for i := 0; i <= 10; i++ {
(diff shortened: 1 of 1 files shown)


functest/deposit_transfer_test.go
```diff
@@ -48,7 +48,15 @@ func TestDepositTransferToken(t *testing.T) {
 	}
 	require.NoError(t, err)
 
-	finalBalance := getBalanceNoErr(t, member, member.ref)
+	var finalBalance *big.Int
```
LinnikD: use waitForFunction from utils_test

megge-dream: i think its better to use `checkBalanceFewTimes` (like at `send_money_test.go`)

PR #2474, opened 2019-08-14
Ins 3243 start launchnet from any test dir

pavel-moskovkin:
**- What I did**
move methods related to launchnet from package functest to a separate package.
this is necessary in order to have an ability of running go tests from other packages, not only from functest dir.
**- How I did it**

**- How to verify it**
run functests


functest/main_test_wrapper.go
```diff
@@ -454,7 +485,7 @@ func teardown() {
 	fmt.Println("[ teardown ] directory for contracts cache was successfully deleted")
 }
 
-func testMainWrapper(m *testing.M) int {
+func TestsMainWrapper(m *testing.M) int {
```
nordicdyno: why you make it public? it also leads this test should have been run by `go test`

functest/test_utils.go
```diff
@@ -44,12 +44,12 @@ import (
 	"github.com/pkg/errors"
 )
 
-type contractInfo struct {
+type ContractInfo struct {
```
nordicdyno: Why you make it public? I don't see any reason.

functest/main_test_wrapper.go
pavel-moskovkin: the reason is to make this method to be called from other directories with test classes, not only from 'functest'. This way other tests could be runned in launchnet as well. QA already have other tests in separate branch

functest/test_utils.go
pavel-moskovkin: utils_test.go used to be test class, though it has no any test methods. I renamed it in order to make its methods and variables public, this change makes  ContractInfo visible from main_test_wrapper.go

functest/main_test_wrapper.go
Tsovak: @pavel-moskovkin  we need to extract the method body to another function if it needs, and use this function here. Just revert these changes, please

functest/test_utils.go
Tsovak: Just revert these changes, please

functest/main_test_wrapper.go
```diff
@@ -246,15 +273,22 @@ func waitForNet() error {
...
 	cwd, err := os.Getwd()
 	if err != nil {
 		return errors.Wrap(err, "[ startNet ] Can't get current working directory")
 	}
 	defer os.Chdir(cwd)
 
-	err = os.Chdir("../")
-	if err != nil {
-		return errors.Wrap(err, "[ startNet  ] Can't change dir")
+	for cwd[len(cwd)-15:] != "insolar/insolar" {
```
nordicdyno: it wouldn't work with go modules, because project's base/working dir could be in path without `insolar/insolar`

testutils/launchnet/launchnet.go
```diff
@@ -134,27 +152,39 @@ func loadMemberKeys(keysPath string, member *user) error {
...
+	if err != nil {
+		return err
+	}
+	for i, md := range MigrationDaemons {
+		path, err := launchnetPath("configs", "migration_daemon_"+strconv.Itoa(i)+"_member_keys.json")
 		if err != nil {
 			return err
 		}
-		migrationDaemons[i] = md
+		err = loadMemberKeys(path, &md)
```
golangcibot: Using a reference for the variable on range scope `md` (from `scopelint`)

testutils/launchnet/launchnet.go
```diff
@@ -58,41 +37,96 @@ var stdin io.WriteCloser
...
+	if err != nil {
+		fmt.Println("error while setup, skip tests: ", err)
+		return 1
+	}
+
+	c := make(chan os.Signal)
+	signal.Notify(c, os.Interrupt)
+
+	go func() {
+		select {
```
golangcibot: S1000: should use a simple channel send/receive instead of `select` with a single case (from `gosimple`)

functest/main_test_wrapper.go
pavel-moskovkin: at the moment this change is necessary, agreed after discussion by @nordicdyno

testutils/launchnet/launchnet.go
```diff
@@ -134,27 +168,40 @@ func loadMemberKeys(keysPath string, member *user) error {
...
 		return err
 	}
-	err = loadMemberKeys(insolarMigrationAdminMemberKeysPath, &migrationAdmin)
+	err = loadMemberKeys(path, &MigrationAdmin)
 	if err != nil {
 		return err
 	}
-	for i, md := range migrationDaemons {
-		err = loadMemberKeys(launchnetPath("configs", "migration_daemon_"+strconv.Itoa(i)+"_member_keys.json"), &md)
+	for i, _ := range MigrationDaemons {
```
golangcibot: should omit 2nd value from range; this loop is equivalent to `for i := range ...` (from `golint`)

testutils/launchnet/launchnet.go
```diff
@@ -58,41 +53,94 @@ var stdin io.WriteCloser
...
-
-type user struct {
-	ref     string
-	privKey string
-	pubKey  string
+	return filepath.Join(parts...), nil
 }
 
-func getNumberNodes() (int, error) {
+func GetNumberNodes() (int, error) {
```
LinnikD: GetNumberNodes -> GetNodesCount

PR #2477, opened 2019-08-14
Ins 3201 fix index rollback

ivanshibitov:
**- What I did**
1. fix incorrect index truncate

**- How I did it**

**- How to verify it**
Run test


ledger/heavy/executor/db_rollback.go
```diff
@@ -67,6 +68,12 @@ func (d *DBRollback) Start(ctx context.Context) error {
 	}
 
 	for idx, db := range d.dbs {
+		switch concretDB := db.(type) {
```
golangcibot: singleCaseSwitch: should rewrite switch statement to if statement (from `gocritic`)

ledger/heavy/executor/finalize_pulse.go
```diff
@@ -22,10 +22,13 @@ import (
 	"github.com/insolar/insolar/insolar"
 	"github.com/insolar/insolar/insolar/pulse"
 	"github.com/insolar/insolar/instrumentation/inslogger"
+	"github.com/insolar/insolar/ledger/object"
 )
 
+var count = 0
```
golangcibot: `count` is unused (from `deadcode`)

ledger/object/index_db.go
```diff
@@ -90,8 +90,29 @@ func (i *IndexDB) SetIndex(ctx context.Context, pn insolar.PulseNumber, bucket r
...
+func (i *IndexDB) UpdateLastKnownPulse(ctx context.Context, topSyncPulse insolar.PulseNumber) error {
+	i.lock.Lock()
+	defer i.lock.Unlock()
+
+	indeces := i.ForPulse(ctx, topSyncPulse)
+	if indeces == nil {
+		return errors.New("No indices for top sync pulse: " + topSyncPulse.String())
+	}
+
+	for idx, _ := range indeces {
```
golangcibot: should omit 2nd value from range; this loop is equivalent to `for idx := range ...` (from `golint`)

ledger/object/index.go
```diff
@@ -31,6 +31,8 @@ import (
 type IndexModifier interface {
 	// SetIndex adds a bucket with provided pulseNumber and ID
 	SetIndex(ctx context.Context, pn insolar.PulseNumber, index record.Index) error
+	// UpdateLastKnownPulse updates last know pulse to given one for all objects from this pulse
+	UpdateLastKnownPulse(ctx context.Context, pn insolar.PulseNumber) error
```
nordicdyno: why it's here? don't abuse interface please

ledger/object/index_db.go
nordicdyno: also use `indexes` word

ledger/object/index.go
andreyromancev: We should probably create another interface for that.

ledger/object/index_db.go
```diff
@@ -143,7 +164,21 @@ func (i *IndexDB) ForID(ctx context.Context, pn insolar.PulseNumber, objID insol
...
-	panic("implement me")
+	indexes := make([]record.Index, 0)
+
+	key := &indexKey{objID: insolar.ID{}, pn: pn}
+	it := i.db.NewIterator(key, false)
+	defer it.Close()
+
+	for it.Next() {
+		key := newIndexKey(it.Key())
+		index, err := i.getBucket(key.pn, key.objID)
```
andreyromancev: it.Value already contains index. No need to do call getBucket.

ledger/object/index.go
ivanshibitov: Interface is called IndexModifier. This method suits to that. No?

insolar/jet/tree.go
```diff
@@ -183,6 +183,7 @@ func getBit(value []byte, index uint8) bool {
 // Panics if index is out of range (value size * 8).
 func setBit(value []byte, index uint8) {
 	if uint(index) >= uint(len(value)*8) {
+		// Usually, it's because of you try to split too deep
```
andreyromancev: `setBit` doesn't know about splits. We should put this message somewhere else.

ivanshibitov: let's decide what we should do with this function

ivanshibitov: I removed it

PR #2480, opened 2019-08-15
INS-3122: make ID and Reference immutable

ita-sammann:
**- What I did**
Deleted all methods allowing to change IDs and References.


ledger/object/lock.go
```diff
@@ -53,13 +53,13 @@ func NewIndexLocker() IndexLocker {
 // If mutex does not exist, it will be created in concurrent safe fashion.
 func (l *idLocker) Lock(id insolar.ID) {
 	// Reset pulse. It should not be considered when locking.
-	id.SetPulse(0)
+	zeroID := *insolar.NewID(0, id.Hash())
```
nordicdyno: I think prefix `zero` is ineligible here. Actually it's not 'zero'-ID, it like ID without pulse prefix, it's better to call it like `normID` or smth like that

PR #2484, opened 2019-08-15
NOISSUE: add ability for multiple calls of Stop/Done

bigbes:
- no deadlocking in Stop
- no panicks in Done

### DIFF
--- a/logicrunner/shutdown/flag.go
+++ b/logicrunner/shutdown/flag.go
@@ -32,16 +32,17 @@ type Flag interface {
 }
 
 type flag struct {
-	stopLock    sync.Mutex
+	stopLock  sync.Mutex
+	isStopped bool
+
 	stopChannel chan struct{}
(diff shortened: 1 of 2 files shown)


logicrunner/shutdown/flag.go
```diff
@@ -32,16 +32,17 @@ type Flag interface {
...
 	stopChannel chan struct{}
-	isStopped   bool
 }
 
 func NewFlag() Flag {
 	return &flag{
 		stopLock:    sync.Mutex{},
 		isStopped:   false,
-		stopChannel: make(chan struct{}, 1),
+		stopChannel: make(chan struct{}, 0),
```
golangcibot: S1019: should use make(chan struct{}) instead (from `gosimple`)

bigbes: fixed

PR #2495, opened 2019-08-16
cache for pulse.DB on heavy. fix hot path

egorikas:
**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/insolar/pulse/db.go
+++ b/insolar/pulse/db.go
@@ -28,8 +28,10 @@ import (
 
 // DB is a DB storage implementation. It saves pulses to disk and does not allow removal.
 type DB struct {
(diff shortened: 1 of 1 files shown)


insolar/pulse/db.go
```diff
@@ -28,8 +28,10 @@ import (
 
 // DB is a DB storage implementation. It saves pulses to disk and does not allow removal.
 type DB struct {
-	db   store.DB
-	lock sync.RWMutex
+	db store.DB
+
+	latest *insolar.Pulse
+	lock   sync.RWMutex
```
andreyromancev: Lock should be above variables it protects.

PR #2503, opened 2019-08-19
NOISSUE: mergeable.yml added

ita-sammann:
Added mergable.yml


ita-sammann:
Force merging because CI is temporary down.

PR #2519, opened 2019-08-20
INS-3287: move slow tests under slowtest tag

nordicdyno:
move slow tests under `slowtest` tag


andreyromancev:
Only tests were added, build failed for known reasons.

PR #2528, opened 2019-08-21
INS-3219: changes in results matcher

ruz:
* delete(rm.unwantedResponses, reqRef) in correct place when
  recieved still executing message, not when we recieved unwanted
  results
* improve logging
* log all not matched unwanted results, this means still executing
  didn't come during the pulse
* use less pointers

### DIFF
--- a/logicrunner/result_matcher.go
+++ b/logicrunner/result_matcher.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"sync"
 
+	"github.com/pkg/errors"
+
 	"github.com/insolar/insolar/insolar"
 	"github.com/insolar/insolar/insolar/flow"
 	"github.com/insolar/insolar/insolar/message"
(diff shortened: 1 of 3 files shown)


logicrunner/result_matcher.go
```diff
@@ -127,6 +135,11 @@ func (rm *resultsMatcher) isStillExecutor(ctx context.Context, object insolar.ID
 func (rm *resultsMatcher) Clear() {
 	rm.lock.Lock()
 	defer rm.lock.Unlock()
+
 	rm.executionNodes = make(map[insolar.Reference]insolar.Reference)
+
+	for reqRef := range rm.unwantedResponses {
+		log.Warn("not claimed response to request ", reqRef.String(), ", not confirmed pending?")
```
megge-dream: Use `inslogger`

ruz: no context to use inslogger :)

ruz: and I'm using, but it's global :(

megge-dream: May be add context to params? It will show pulse number, it can be useful

PR #2542, opened 2019-08-22
INS-3219: when execution returns to node executing pending

ruz:
old code was not accounting existing of registry and that broker
is new every pulse.

new reality is that object is in pending even if pending execution
is on this node

old code was starting second mutable execution on the same node
and if it was the same request then context on outgoing calls was
shared resulting in different Nonce values and creating new
requests.


ita-sammann:
Unit tests passed. Force merging.

PR #2544, opened 2019-08-22
NOISSUE: Move SyncT to testutuils

Vany:
make code more consistent

### DIFF
--- a/functest/pressure_test.go
+++ b/functest/pressure_test.go
@@ -22,9 +22,9 @@ import (
 	"sync"
 	"testing"
 
-	"github.com/insolar/insolar/testutils"
-
 	"github.com/stretchr/testify/require"
+
+	"github.com/insolar/insolar/testutils"
 )
 
 func TestPressureOnSystem(t *testing.T) {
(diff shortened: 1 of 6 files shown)


functest/pressure_test.go
```diff
@@ -22,9 +22,9 @@ import (
 	"sync"
 	"testing"
 
-	"github.com/stretchr/testify/require"
+	"github.com/insolar/insolar/testutils"
 
-	"github.com/insolar/insolar/insolar/utils"
+	"github.com/stretchr/testify/require"
```
aalekseev-ru: we have package ordering rules - native, third party, insolar

logicrunner/shutdown/flag_test.go
```diff
@@ -20,15 +20,16 @@ import (
 	"testing"
 	"time"
 
+	"github.com/insolar/insolar/testutils"
```
aalekseev-ru: same here

ita-sammann:
Force merged. Unit tests passed but could not notify GitHub.

PR #2549, opened 2019-08-22
Ins 3295 additional checks for integration tests

aMFeriuSS:
**- What I did**
- Move `./ledger/light/integration/` tests to `slowtest`-tag
- Remove useless `Server.SetPulse` into tests initialisation (when if don't needed)
- Add new check for `request from another object with root object reason in detached mode when reason closed already.`

**- How to verify it**
`go test -tags slowtest ./ledger/light/integration/`

### DIFF
--- a/ledger/light/integration/operations_test.go
+++ b/ledger/light/integration/operations_test.go
@@ -13,6 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
+// +build slowtest
 
 package integration_test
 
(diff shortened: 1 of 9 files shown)


ledger/light/integration/operations_test.go
```diff
@@ -107,6 +110,36 @@ func MakeSetIncomingRequest(objectID, reasonID insolar.ID, reasonObjectID insola
 	return pl, rec
 }
 
+func MakeSetIncomingRequestDetached(
+	objectID insolar.ID,
+	reasonID insolar.ID,
+	reasonObjectID insolar.ID,
+	isCreation bool,
```
andreyromancev: Can never be creation.

aMFeriuSS: Ok, done

PR #2551, opened 2019-08-22
INS-3219: execution broker has no all knowledge anymore

ruz:
deduplicate using registry as well, pulse ends - new broker, never
put into queue what we're executing right now


ita-sammann:
Unit tests passed. Force merging

PR #2560, opened 2019-08-23
#INS-3293 fix goroutine leak in test

SergeyRadist:
**- What I did**
added force terminate in go-actors 
**- How I did it**

**- How to verify it**

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -262,7 +262,7 @@
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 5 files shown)


bigbes:
Are you sure fortytw2/leaktest will work well with parallel tests and multi-package testing?

bigbes:
What about other places where system is created (Init function in LR) and not closed?

SergeyRadist:
> Are you sure fortytw2/leaktest will work well with parallel tests and multi-package testing?
no, it won't thx


> What about other places where system is created (Init function in LR) and not closed?
added in logicrunner.stop

PR #2576, opened 2019-08-26
WLT-928: fix bug with check amount

sansaian:
**- What I did**
fixes https://insolar.atlassian.net/browse/WLT-928
**- How I did it**
I’m only checking active demons whose amount is not equal empty string
**- How to verify it**
starting new func test
change contract deposit and change func_tests


functest/test_utils.go
```diff
@@ -46,6 +46,11 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	countTreeActiveDaemon = iota + 3
```
sexyhamster: Three?

sansaian: misprint

functest/test_utils.go
```diff
@@ -249,15 +254,17 @@ func getStatus(t testing.TB) statusResponse {
 	return rpcStatusResponse.Result
 }
 
-func activateDaemons(t *testing.T) {
+func activateDaemons(t *testing.T, countDaemon int) {
```
sexyhamster: actually you can return list of migrationDaemons, so you will iterate over this list instead of iterating `count` times

that might look better

logicrunner/builtin/contract/deposit/deposit.go
```diff
@@ -131,17 +129,24 @@ func (d *Deposit) Confirm(migrationDaemonRef string, txHash string, amountStr st
...
+		return fmt.Errorf(" list with migration daemons member is empty ")
+	}
+	result := ""
+	for _, migrationRef := range activeDaemons {
+		if amount, ok := d.MigrationDaemonConfirms[migrationRef]; ok {
+			if result == "" {
+				result = amount
+				continue
+			}
+			if result != d.MigrationDaemonConfirms[migrationRef] {
```
megge-dream: Why not just use `amount ` instead of `d.MigrationDaemonConfirms[migrationRef]?`

sansaian: Done

functest/test_utils.go
sansaian: Done. If I understood correctly

sexyhamster: hmm
my idea was :
`func activateDaemons(t *testing.T, countDaemon int) []*User`
so you don't actually need global list of Users. You can create exact ammount for each test

but I'll leave the decision for @megge-dream I'm not sure about code style

sexyhamster: To explain myself, with 
`func activateDaemons(t *testing.T, countDaemon int) []*User`
you could write tests like this
`
for _, daemon := activateDaemons(t, 3) {
   // do stuff with daemon
}
`

PR #2579, opened 2019-08-26
WLT-907: move deposits from member to wallet

pavlushev:
What I did:
- moved deposits from member to wallet


logicrunner/builtin/contract/wallet/wallet.go
```diff
@@ -145,3 +148,45 @@ func (w *Wallet) Accept(amountStr string, assetName string) error {
 	acc := account.GetObject(*accRef)
 	return acc.Accept(amountStr)
 }
+
+// SetDeposit method stores deposit reference in member it belongs to
```
megge-dream: Name of func is `AddDeposit `, not `SetDeposit `

logicrunner/builtin/contract/wallet/wallet.go
```diff
@@ -145,3 +148,45 @@ func (w *Wallet) Accept(amountStr string, assetName string) error {
...
+
+// FindDeposit finds deposit for this wallet with this transaction hash.
+// ins:immutable
+func (w *Wallet) FindDeposit(transactionHash string) (bool, *insolar.Reference, error) {
+	if depositReferenceStr, ok := w.Deposits[transactionHash]; ok {
+		depositReference, _ := insolar.NewReferenceFromBase58(depositReferenceStr)
+		return true, depositReference, nil
+	}
+
+	return false, insolar.NewEmptyReference(), nil
```
megge-dream: Why not ` return false, nil, nil`?

PR #2586, opened 2019-08-27
INS-3363: fix fatal and rewrite tests

LinnikD:
fix fatal and rewrite tests for it


ledger/heavy/executor/initial_state_keeper.go
```diff
@@ -170,7 +180,12 @@ func (isk *InitialStateKeeper) Get(ctx context.Context, lightExecutor insolar.Re
 
 		if light.Equal(lightExecutor) {
 			jetIDs = append(jetIDs, id)
-			drops = append(drops, jetDrop)
+
+			parentID := jet.Parent(id)
```
ivanshibitov: Looks like incorrect logic if there was no split

insolar/jet/tree.go
```diff
@@ -99,6 +99,14 @@ func (j *Jet) ExtractLeafIDs(ids *[]insolar.JetID, path []byte, depth uint8) {
 	}
 }
 
+//go:generate minimock -i github.com/insolar/insolar/insolar/jet.TreeManager -o ./ -s _mock.go -g
+
+// TreeManager interface for operations with JetTree
+type TreeManager interface {
```
andreyromancev: It's not a manger. Maybe something like `TreeDescriptor`?

ledger/heavy/executor/initial_state_keeper.go
```diff
@@ -73,7 +76,9 @@ func NewInitialStateKeeper(
 		indexStorage:          indexStorage,
 		dropStorage:           dropStorage,
 		syncPulse:             jetKeeper.TopSyncPulse(),
+		jetTree:               jet.NewTree(true),
```
andreyromancev: `StateKeeper` is a singleton. This field is a temporary object to perform calculations. It might produce serious and hard to find bugs.

ledger/heavy/executor/initial_state_keeper.go
```diff
@@ -162,6 +172,10 @@ func (isk *InitialStateKeeper) Get(ctx context.Context, lightExecutor insolar.Re
 	indexes := make([]record.Index, 0)
 
 	logger.Debugf("[ InitialStateKeeper ] Getting drops for: %s in pulse: %s", lightExecutor.String(), pulse.String())
+
+	// Exclude sending two equal drops to single LME after split
+	excludeDrops := make(map[insolar.JetID]bool)
```
andreyromancev: `excludeDrops` is confusing. We are not excluding them, just deduplicating. Something like `existingDrops` might be a better option. 
Also, we are only using it as a set. A common convention to do that is to use a map of empty structs.

PR #2600, opened 2019-08-28
Ins 2871 more unit tests for ledger

aMFeriuSS:
**- What I did**
Improve unit tests for `set_outgoing_request` and `set_result` handlers


ledger/light/handle/set_outgoing_request.go
```diff
@@ -42,20 +42,25 @@ func NewSetOutgoingRequest(dep *proc.Dependencies, msg payload.Meta, passed bool
...
+	rec := record.Unwrap(&virtual)
 	request, ok := rec.(*record.OutgoingRequest)
 	if !ok {
-		return fmt.Errorf("SetOutgoingRequest.Present: wrong request type: %T", rec)
+		return fmt.Errorf("wrong request type: %T", rec)
 	}
 
-	if request.AffinityRef() == nil {
-		return errors.New("SetOutgoingRequest.Present: object is nil")
+	objectID := *request.AffinityRef().Record()
```
andreyromancev: nil pointer dereference. Better use `request.AffinityRef() == nil || *request.AffinityRef().Record().IsEmpty()`

aMFeriuSS: We have non-reference field Caller in OutgoingRequest struct (Caller is returned when we call AffinityRef method on request). So we cant't get nil reference here.

PR #2610, opened 2019-08-28
INS-3274: virtual integration tests

sexyhamster:
**- What I did**
virtual integration skeleton

**- How I did it**
I've used light integration as an example. But logicrunner is more complex, so it took a while.

For now network with only one light and mocked light could be built.

TODO: be able to start two interconnected servers, to be able to process more complex cases

**- How to verify it**
run happy path test


logicrunner/artifacts/instrument.go
```diff
@@ -18,9 +18,10 @@ package artifacts
 
 import (
 	"context"
-	"github.com/insolar/insolar/insolar/flow"
 	"time"
 
+	"github.com/insolar/insolar/insolar/flow"
+
```
aalekseev-ru: strange changes, if you want to do it properly - first block - inner packages, second block - outer packages, third . block - insolar packages

virtual/integration/virtual_test.go
```diff
@@ -0,0 +1,180 @@
...
+	"github.com/insolar/insolar/insolar/record"
+	"github.com/insolar/insolar/insolar/reply"
+	"github.com/insolar/insolar/instrumentation/inslogger"
+	"github.com/insolar/insolar/logicrunner/machinesmanager"
+	"github.com/insolar/insolar/platformpolicy"
+	"github.com/insolar/insolar/testutils"
+)
+
+func TestVirtual_BasicOperations(t *testing.T) {
+	t.Parallel()
```
aalekseev-ru: if I remember it right we remove t.Parallel() because it causes debug problems

sexyhamster: well here it should be no problem (and it is used in light)

sexyhamster:
reverted Parcel part.
Will do it in separate PR

virtual/integration/virtual_test.go
sexyhamster: removed t.Parallel

PR #2621, opened 2019-08-29
INS-3387: Remove ObjectRequestID from GetObject and Memory from State.

andreyromancev:
Removed unused fields from proto files.


insolar/record/record.proto
```diff
@@ -107,22 +91,11 @@ message Result {
...
-    bytes TypeDeclaration = 22;
-}
-
 message Code {
     option (gogoproto.face) = true;
 
     int32 polymorph = 16;
 
-    bytes Domain = 20 [(gogoproto.customtype) = "github.com/insolar/insolar/insolar.Reference", (gogoproto.nullable) = false];
     bytes Request = 21 [(gogoproto.customtype) = "github.com/insolar/insolar/insolar.Reference", (gogoproto.nullable) = false];
```
aMFeriuSS: Here and below we should start field numbers from 20

insolar/record/record.proto
```diff
@@ -191,12 +161,9 @@ message Virtual {
 
     oneof union {
         Genesis Genesis = 101;
-        Child Child = 102;
-        Jet Jet = 103;
         IncomingRequest IncomingRequest = 104;
```
aMFeriuSS: Check fieldnumbers here

insolar/record/record.proto
```diff
@@ -191,12 +161,9 @@ message Virtual {
...
         OutgoingRequest OutgoingRequest = 105;
         Result Result = 106;
-        Type Type = 107;
         Code Code = 108;
```
aMFeriuSS: And here

PR #2625, opened 2019-08-29
WLT-950: fix bug with load migration address

pavlushev:
What I did:
- added trimAddress
- added trimmedAddress to slice

### DIFF
--- a/functest/deposit_transfer_test.go
+++ b/functest/deposit_transfer_test.go
@@ -73,7 +73,7 @@ func TestDepositTransferToken(t *testing.T) {
 	_, err = waitUntilRequestProcessed(anon, time.Second*20, time.Second, 20)
 	require.NoError(t, err)
 	if err == nil {
-		fmt.Print("SUCCESS")
+		fmt.Print("SUCCESS\n")
 	}
 	checkBalanceFewTimes(t, member, member.Ref, secondBalance)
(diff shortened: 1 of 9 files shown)


functest/deposit_transfer_test.go
```diff
@@ -54,21 +54,27 @@ func TestDepositTransferToken(t *testing.T) {
 	}
 	_, err := waitUntilRequestProcessed(anon, time.Second*20, time.Second, 20)
 	require.NoError(t, err)
+	if err == nil {
```
megge-dream: if err != nil test wiil fail on previous line

functest/deposit_transfer_test.go
```diff
@@ -54,21 +54,27 @@ func TestDepositTransferToken(t *testing.T) {
...
 		}
-		fmt.Println("Vesting STEP: --- RETURNED   " + err.Error())
+		fmt.Print(".")
 		return api.CallMethodReply{
 			Error: &foundation.Error{err.Error()},
 		}
 	}
 	_, err = waitUntilRequestProcessed(anon, time.Second*20, time.Second, 20)
 	require.NoError(t, err)
+	if err == nil {
```
megge-dream: if err != nil test wiil fail on previous line

PR #2632, opened 2019-08-29
Readme update

sofiayem:
**- What we did**

Completely reworked README.


ita-sammann:
Only readme is affected. Force merging.

PR #2649, opened 2019-09-02
INS-3321: index not found should be final error

sexyhamster:
**- What I did**
ensured index not found as logical error

**- How I did it**
with helper

**- How to verify it**
run tests

### DIFF
--- a/logicrunner/procs.go
+++ b/logicrunner/procs.go
@@ -146,3 +146,10 @@ func (r *RecordErrorResult) Proceed(ctx context.Context) error {
 
 	return nil
 }
+
+func ProcessLogicalError(err error) (error, bool) {
+	if e, ok := errors.Cause(err).(*payload.CodedError); ok && e.Code == payload.CodeNotFound {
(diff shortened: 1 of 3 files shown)


logicrunner/procs.go
```diff
@@ -146,3 +146,10 @@ func (r *RecordErrorResult) Proceed(ctx context.Context) error {
 
 	return nil
 }
+
+func ProcessLogicalError(err error) (error, bool) {
```
golangcibot: error should be the last type when returning multiple items (from `golint`)

logicrunner/procs.go
```diff
@@ -146,3 +146,10 @@ func (r *RecordErrorResult) Proceed(ctx context.Context) error {
...
+	if e, ok := errors.Cause(err).(*payload.CodedError); ok && e.Code == payload.CodeNotFound {
+		return errors.New("not found"), true
+	}
+	return err, false
```
andreyromancev: Why return the same error? Maybe just return `bool`?

PR #2651, opened 2019-09-02
INS-3122: replace some ObjectReferences with RecordReferences

bigbes:
-- Record References are references with zeroed base part (and ends with .record in textual repr)
-- Object References (Local) have left and right part duplicated (and starts with 1 in textual repr)

* Replace Code/Reason/Request references with RecordReferences (all other references should be ObjectReferences)
* Add checks for ObjectReferences/RecordReferences in incoming messages
* Actualize unit+func tests

### DIFF
--- a/insolar/payload/payload.go
+++ b/insolar/payload/payload.go
@@ -20,6 +20,8 @@ import (
 	"github.com/gogo/protobuf/proto"
 	base58 "github.com/jbenet/go-base58"
 	"github.com/pkg/errors"
+
+	"github.com/insolar/insolar/insolar"
 )
 
 type Type uint32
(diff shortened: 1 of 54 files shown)


insolar/payload/payload.go
```diff
@@ -139,6 +141,14 @@ func (h *MessageHash) IsZero() bool {
 	return true
 }
 
+func (m RequestInfo) RequestAsObjectReference() *insolar.Reference {
```
andreyromancev: Can we move these to logicrunner helpers? Doesn't have to be a method.

bigbes: Both of these functions? `RequestAsObjectReference`+`RequestReference`?

andreyromancev: Yep. Let's not pollute the record package.

bigbes: kk, pushed

insolar/payload/payload.go
```diff
@@ -139,6 +141,10 @@ func (h *MessageHash) IsZero() bool {
 	return true
 }
 
+func (m RequestInfo) RequestReference() *insolar.Reference {
```
andreyromancev: Same here, let's move this somewhere else.

bigbes: Done; I've rewritten code, but forgot to remove old wrapper.

PR #2654, opened 2019-09-02
INS-3293 / INS-3371: stop OutgoingRequestSender

afiskon:
How to  test:

```
$ INSOLAR_LOG_LEVEL=warn go test ./logicrunner -run '^TestLogicRunner/TestConcurrency$' -count=10000 -race -v &>2
$ cat 2 | grep 'go-actors/actor/mailbox/mailbox.go' | wc -l
0
```

### DIFF
--- a/logicrunner/outgoingsender.go
+++ b/logicrunner/outgoingsender.go
@@ -8,6 +8,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/insolar/go-actors/actor"
+	aerr "github.com/insolar/go-actors/actor/errors"
 
 	"github.com/insolar/insolar/insolar"
 	"github.com/insolar/insolar/insolar/payload"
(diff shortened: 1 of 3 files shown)


logicrunner/outgoingsender.go
```diff
@@ -109,9 +114,19 @@ func (rs *outgoingRequestSender) SendAbandonedOutgoingRequest(ctx context.Contex
...
-func (rs *outgoingRequestSender) Stop(_ context.Context) {
-	// not implemented yet, see INS-3293, INS-3371
-	// rs.as.CloseAll()
+func (rs *outgoingRequestSender) Stop(ctx context.Context) {
+	resultChan := make(chan struct{}, 1)
+	msg := stopOutgoingRequestSenderMessage{
+		resultChan: resultChan,
+	}
+	err := rs.as.SendPriority(rs.senderPid, msg)
+	if err != nil { // this is unlikely to happen
```
sexyhamster: for unlikely use panic to be able to backtrace

afiskon: I don't think I agree. IIRC there is no such rule in our code style. Also the only scenario when SendPriority can fail is when actor we send a message to doesn't exist.

PR #2656, opened 2019-09-03
Ins 3418 default configs generator

egorikas:
**- What I did**

**- How I did it**

**- How to verify it**


configuration/export.go
```diff
@@ -17,6 +17,8 @@
 package configuration
 
 // Exporter holds exporter configuration.
+// Is was assumed, that exporter will be used for exporting data for observer
```
andreyromancev: "It"

PR #2680, opened 2019-09-04
Ins 3050 functest logs rotate

nordicdyno:
* add tool for log rotation
* rotate logs on every functest start


opencensus-race/opencensus_race.go
```diff
@@ -0,0 +1,76 @@
+package opencensus_race
```
golangcibot: don't use an underscore in package name (from `golint`)

PR #2681, opened 2019-09-04
NOISSUE: added changelog generator script

ita-sammann:
Added changelog generator script.
Requires Python 3, Jinja 2, requests, logzero.
```
pip3 install jinja2 requests logzero
``` 

Usage:
```
scripts/changelog.py v0.9.0-rc3 v0.9.0-rc4 -e <your email> -t <JIRA api token> --html=.artifacts/release.html
```

You can generate the token here: https://id.atlassian.com/manage/api-tokens

### DIFF
--- a/scripts/changelog.py
+++ b/scripts/changelog.py
@@ -37,11 +37,8 @@ def main(args):
         print_changelog(args.current_release, args.new_release, issues)
 
 
+# git log v0.9.0-rc3..v0.9.0-rc4 --oneline | grep -E '[A-Z]+-\d+' | perl -ne '/([A-Z]+-\d+)/; print "$1\n"' | sort -u
 def get_issues_from_git(current_release, new_release):
-    # git log v0.9.0-rc3..v0.9.0-rc4 --oneline
-    #   | grep -E '[A-Z]+-\d+'
-    #   | perl -ne '/([A-Z]+-\d+)/; print "$1\n"'
-    #   | sort -u
     git_log = subprocess.run(
         ['git', 'log', f'{current_release}..{new_release}', '--oneline'],
         capture_output=True,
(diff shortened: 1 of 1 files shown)


scripts/changelog.py
```diff
@@ -0,0 +1,223 @@
...
+    file.write(html)
+    file.close()
+
+
+def render_html(context):
+    tpl = jinja2.Environment().from_string(page_tpl())
+    return tpl.render(context)
+
+
+def page_tpl():
```
dmitry-panchenko: Why do you use function here? Make it constant

scripts/changelog.py
```diff
@@ -0,0 +1,223 @@
...
+        key = i['fields']['status']['statusCategory']['key']
+        if key in status_colors:
+            i['fields']['status']['statusCategory']['status'] = status_colors[key]
+    return context
+
+
+def save_html(context, filename):
+    context = prepare_context(context)
+    html = render_html(context)
+    file = open(filename, 'w+')
```
dmitry-panchenko: `with` statement?

scripts/changelog.py
```diff
@@ -0,0 +1,223 @@
...
+            issues.append(r.json())
+        except ValueError:
+            logger.error(f'Failed to get info for issue {key}')
+    return issues
+
+
+def print_changelog(current, new, issues):
+    print(f'Changelog from {current} to {new}:')
+    for i in issues:
+        issue = i
```
dmitry-panchenko: Why not just `for issue in issues `?

scripts/changelog.py
```diff
@@ -0,0 +1,223 @@
...
+"""
+
+import argparse
+import re
+import subprocess
+import sys
+
+import jinja2
+import requests
+from logzero import logger
```
dmitry-panchenko:
```
import logging

logger = logging.getLogger(__name__)
```

scripts/changelog.py
ita-sammann: ✓

scripts/changelog.py
ita-sammann: ✓

scripts/changelog.py
ita-sammann: ✓

scripts/changelog.py
ita-sammann: Nope :-P

PR #2683, opened 2019-09-04
INS-3465: return err.Error() in json2.Error.Message

afiskon:
Return err.Error() in json2.Error.Message, otherwise the real failure is hidden during execution of tests.


bigbes:
Probably it should be reviewed by someone who made this API.

I'd prefer to concatenate it on client side with extended error data.

PR #2692, opened 2019-09-05
NOISSUE: Ticking every second and dumping log ever faster

Vany:
NOISSUE: Ticking every second and dumping log ever faster

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -107,7 +107,6 @@ var GlobalLogger = func() insolar.Logger {
 
 	if logger.Is(insolar.DebugLevel) {
 		go func() {
-			logger.Debug("Start logger ticker")
 			for {
 				time.Sleep(time.Second - time.Now().Sub(time.Now().Truncate(time.Second)))
 				logger.Debug("Logger tick")
--- a/log/log.go
+++ b/log/log.go
@@ -108,7 +108,8 @@ var GlobalLogger = func() insolar.Logger {
 	if logger.Is(insolar.DebugLevel) {
 		go func() {
 			for {
-				time.Sleep(time.Second - time.Now().Sub(time.Now().Truncate(time.Second)))
+				// Tick between seconds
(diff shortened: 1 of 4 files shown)


log/log.go
```diff
@@ -104,6 +105,13 @@ var GlobalLogger = func() insolar.Logger {
 		panic("couldn't initialize global logger with default config")
 	}
 
+	go func() {
+		for {
+			time.Sleep(time.Now().Sub(time.Now().Truncate(time.Second)))
```
golangcibot: S1012: should use `time.Since` instead of `time.Now().Sub` (from `gosimple`)

log/log.go
```diff
@@ -104,6 +105,15 @@ var GlobalLogger = func() insolar.Logger {
 		panic("couldn't initialize global logger with default config")
 	}
 
+	if logger.Is(insolar.DebugLevel) {
+		go func() {
+			for {
+				time.Sleep(time.Second - time.Now().Sub(time.Now().Truncate(time.Second)))
```
golangcibot: S1012: should use `time.Since` instead of `time.Now().Sub` (from `gosimple`)

sexyhamster: ++ it's better to create ticker

Vany: No, ticker does different thing.

log/log.go
```diff
@@ -104,6 +105,16 @@ var GlobalLogger = func() insolar.Logger {
 		panic("couldn't initialize global logger with default config")
 	}
 
+	if logger.Is(insolar.DebugLevel) {
+		go func() {
+			for {
```
sexyhamster: my point was
`
ticker := time.NewTicker(time.Second)
	for range ticker.C {
		fmt.Println("Hello, playground")
	}
`

sexyhamster: it's cleaner and more "go way"

Vany: but it's not that accurate

PR #2695, opened 2019-09-05
NOISSUE: refactor Docker build 

nordicdyno:
* add Dockerfile for base image
* move outdated testnet-node files to sub dir


bigbes:
Why centos?

docker/Dockerfile
```diff
@@ -1,97 +1,53 @@
...
+RUN echo -n "<DISK USAGE> before apt-get: " && du -sh / 2>/dev/null || true
+# tools for functest (launchnet)
+RUN apt-get -y update && apt-get -y install jq lsof psmisc && apt-get clean
 
-ENTRYPOINT [ "/opt/bin/docker-entrypoint.sh" ]
-CMD []
+# adds 350mb (>250mb – binaries)
+RUN echo -n "<DISK USAGE> before insolar files copy: " && du -sh / 2>/dev/null || true
+COPY --from=build /go/src/github.com/insolar/insolar /go/src/github.com/insolar/insolar
+RUN ls -al ./bin && ls -al ./
```
bigbes: why?

docker/Dockerfile
```diff
@@ -1,97 +1,53 @@
...
+# adds 20mb
+RUN echo -n "<DISK USAGE> before apt-get: " && du -sh / 2>/dev/null || true
+# tools for functest (launchnet)
+RUN apt-get -y update && apt-get -y install jq lsof psmisc && apt-get clean
 
-ENTRYPOINT [ "/opt/bin/docker-entrypoint.sh" ]
-CMD []
+# adds 350mb (>250mb – binaries)
+RUN echo -n "<DISK USAGE> before insolar files copy: " && du -sh / 2>/dev/null || true
+COPY --from=build /go/src/github.com/insolar/insolar /go/src/github.com/insolar/insolar
```
bigbes: copy bins only

nordicdyno:
> Why centos?

short answer: for consistency
the loner one: Centos is a default OS for running our binaries, and it's not gonna change in near future

docker/Dockerfile
nordicdyno: nope. this is base image for tests also, we need sources and dependencies here

bigbes:
You've tried to make that image smaller, but in the end it's 1.2GB, it's a lot of wasted space.

nordicdyno:
> You've tried to make that image smaller, but in the end it's 1.2GB, it's a lot of wasted space.

1) it's much smaller
2) it's not "wasted space". we need all this files/bianries for reproducible builds/test till we use insgorund
3) if you need smaller image with only one binary inside, just build it on top of this image (i.e copy file to scratch container)

docker/Dockerfile
nordicdyno: Just to be sure we not forget smth, could be useful in CI builds. I can remove/commenting it.

docker/Dockerfile
```diff
@@ -1,97 +1,53 @@
...
-            git           \
-            make          \
-            alpine-sdk    \
-            bash
-
-COPY [ "/docker/genconfig/genconfig.go", \
-       "/opt/tmp/" ]
-
-COPY . /go/src/github.com/insolar/insolar
+RUN mkdir -p /go/src/github.com/insolar/insolar
```
Tsovak: actually, it's not necessary

docker/testnet-node/insolard.Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM centos:7.6.1810 AS insolard
```
Tsovak: what about alpine?

docker/testnet-node/insgorund.Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM centos:7.6.1810 AS insgorund
```
Tsovak: what about alpine?

nordicdyno: our base OS is centos, in derevative images you could use scratch as base image if you need small image size for running/pulling in production

docker/Dockerfile
nordicdyno: Yeah you are right, removed

PR #2705, opened 2019-09-06
INS-3483: logger waits till metrics will be initialized

Vany:
INS-3483: logger waits till metrics will be initialized

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -105,17 +105,21 @@ var GlobalLogger = func() insolar.Logger {
 		panic("couldn't initialize global logger with default config")
 	}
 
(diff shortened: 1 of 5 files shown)


log/log.go
```diff
@@ -105,17 +105,21 @@ var GlobalLogger = func() insolar.Logger {
 		panic("couldn't initialize global logger with default config")
 	}
 
+	ret := logger.WithCaller(true).WithSkipFrameCount(1).WithField("loginstance", "global_default")
+
 	if logger.Is(insolar.DebugLevel) {
 		go func() {
+			innerLogger := ret.WithCaller(false)
+			time.Sleep(time.Second)
```
andreyromancev: This is not a proper fix.

PR #2731, opened 2019-09-10
NOISSUE: fix bug re bad PendingFinished and StillExecution messages

bigbes:
* when we first send PendingFinished we won't move our state (in some
  situations) to InPending
* when OnPulse happens, sometime, we won't move our state to InPending,
  too
* Added some more verbose logging


logicrunner/executionregistry/execution_registry.go
```diff
@@ -118,7 +118,7 @@ func (r *executionRegistry) OnPulse(_ context.Context) []payload.Payload {
 	messages := make([]payload.Payload, 0)
 	if len(r.registry) != 0 {
 		requestRefs := make([]insolar.Reference, 0, len(r.registry))
-		for requestRef := range r.registry {
+		for requestRef, _ := range r.registry {
```
golangcibot: should omit 2nd value from range; this loop is equivalent to `for requestRef := range ...` (from `golint`)

PR #2778, opened 2019-09-16
INS-3574: not to print CLI help on insolard successful start

nordicdyno:
remove annoying help print

### DIFF
--- a/cmd/insolard/main.go
+++ b/cmd/insolard/main.go
@@ -39,7 +39,11 @@ type inputParams struct {
 }
 
 func parseInputParams() inputParams {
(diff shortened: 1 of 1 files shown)


cmd/insolard/main.go
```diff
@@ -39,7 +39,11 @@ type inputParams struct {
 }
 
 func parseInputParams() inputParams {
-	var rootCmd = &cobra.Command{Use: "insolard"}
```
sexyhamster: please put TODO here, so we 

1. either go with cobra command (put start into Run)
2. remove cobra (because config reading could be done without cobra)

It looks ugly as hell :(

nordicdyno:
fixed in PR #2828

PR #2783, opened 2019-09-16
NOISSUE: prioritize network pulse stream

Vany:
NOISSUE: prioritize network pulse stream

### DIFF
--- a/network/consensus/adapters/handlers.go
+++ b/network/consensus/adapters/handlers.go
@@ -54,6 +54,7 @@ import (
 	"bytes"
 	"context"
 	"io"
+	"runtime"
 
 	"github.com/insolar/insolar/instrumentation/insmetrics"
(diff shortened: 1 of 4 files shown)


Vany:
> Where is the proof, that this increases the speed of execution?
@egorikas  this is an experiment to make this proof.
> 
> > With the Go threading model, calls to C code, assembler code, or blocking system calls occur in the same thread as the calling Go code, which is managed by the Go runtime scheduler.
> > The os.LockOSThread() mechanism is mostly useful when Go has to interface with some foreign library (a C library for instance). It guarantees that several successive calls to this library will be done in the same thread.

Vany:
> Already attempted in various ways. This will NOT work. Please cancel this change.

@cyraxred please provide results of this experiments.

PR #2802, opened 2019-09-18
INS-3569: Add filament metrics.

andreyromancev:
Add filament metrics.

### DIFF
--- a/ledger/light/executor/filaments.go
+++ b/ledger/light/executor/filaments.go
@@ -23,6 +23,7 @@ import (
 	"sync"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


ledger/light/executor/filaments.go
```diff
@@ -573,6 +574,10 @@ func (i *filamentIterator) Prev(ctx context.Context) (record.CompositeFilamentRe
 		return record.CompositeFilamentRecord{}, object.ErrNotFound
 	}
 
+	defer func() {
```
nordicdyno: could be just `defer stats.Record(ctx, statFilamentLength.M(1))`

PR #2804, opened 2019-09-18
INS-3571: limit concurrent immutable execution

bigbes:
Limit concurrent immutable execution queue

### DIFF
--- a/logicrunner/executionbroker.go
+++ b/logicrunner/executionbroker.go
@@ -19,7 +19,6 @@ package logicrunner
 import (
 	"context"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	watermillMsg "github.com/ThreeDotsLabs/watermill/message"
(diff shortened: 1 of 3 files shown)


logicrunner/executionbroker.go
```diff
@@ -196,9 +197,10 @@ func (q *ExecutionBroker) finishTask(ctx context.Context, transcript *common.Tra
 	}
 }
 
-func (q *ExecutionBroker) processTranscript(ctx context.Context, transcript *common.Transcript) {
+func (q *ExecutionBroker) processTranscript(ctx context.Context, state *processorState, transcript *common.Transcript) {
```
golangcibot: `(*ExecutionBroker).processTranscript` - `state` is unused (from `unparam`)

bigbes: fixed

logicrunner/executionbroker.go
```diff
@@ -290,22 +290,36 @@ func (q *ExecutionBroker) IsKnownRequest(ctx context.Context, req insolar.Refere
 
 // startProcessors starts independent processing of mutable and immutable queues
 func (q *ExecutionBroker) startProcessors(ctx context.Context) {
-	q.startProcessor(ctx, &q.immutable)
+	for i := q.startProcessorCount(ctx, &q.immutable); i > 0; i-- {
```
ruz: I don't like this code. startProcessors is called all the time concurrently and old code was guaranteeing that processor will be started only one. New code starts fights around counting queue length, available slots left. Also, this code gets complicated and I feel like it is possible that we wouldn't start a processor when queue is not empty.

PR #2806, opened 2019-09-18
INS-2682: Caller is mutually exclusive with APINode, not Reason

bigbes:
Revert of previous fix


andreyromancev:
Functest passed https://teamcity.insolar.io/viewLog.html?buildId=133681&buildTypeId=AwesomeCi_FuncTests

Jepsen failed due to log processing.

PR #2817, opened 2019-09-20
INS-3584: pass request to AM.client.GetObject

ruz:
moving towards execution order


bigbes:
@ruz @andreyromancev It looks like it's easier to force check of `RequestID`.

The reason we can't do it is alternative logic for `Prototypes`, but that easily fixed with appending next check - if object we trying to fetch is prototype and it's prototype of request passed with RequestID - than allow execution.

Not my call to force you to implement that, but i think that check should be mandatory.

PR #2821, opened 2019-09-20
Ins 3589 disallow side effect on outgoing

SergeyRadist:
**- What I did**
added new check in set_result that disallows side effect on outgoing request
**- How I did it**

**- How to verify it**

### DIFF
--- a/insolar/payload/error_code.go
+++ b/insolar/payload/error_code.go
@@ -26,12 +26,10 @@ const (
 	CodeNoStartPulse = 1005
 
 	CodeRequestNotFound   = 1006
-	CodeInvalidRequest    = 1007
+	CodeRequestInvalid    = 1007
 	CodeReasonNotFound    = 1008
 	CodeReasonIsWrong     = 1009
 	CodeNonClosedOutgoing = 1010
-
-	CodeRequestIsWrong = 1011
 )
 
 type CodedError struct {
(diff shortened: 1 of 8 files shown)


ledger/light/proc/set_result.go
```diff
@@ -148,6 +148,10 @@ func (p *SetResult) Proceed(ctx context.Context) error {
 		return errors.Wrap(err, "failed to calculate pending requests")
 	}
 	closedRequest, err := findClosed(opened, p.result)
+
+	if p.sideEffect != nil {
+		err = checkRequest(closedRequest)
```
bigbes: checkRequestSideEffect?

insolar/payload/error_code.go
```diff
@@ -30,6 +30,8 @@ const (
 	CodeReasonNotFound    = 1008
 	CodeReasonIsWrong     = 1009
 	CodeNonClosedOutgoing = 1010
+
+	CodeRequestIsWrong = 1011
```
andreyromancev: We already have `CodeInvalidRequest`

insolar/record/record.go
```diff
@@ -164,6 +164,8 @@ type Request interface {
 	// IsTemporaryUploadCode tells us that that request is temporary hack
 	// for uploading code.
 	IsTemporaryUploadCode() bool
+	// IsMutable returns mutable-state of request
+	IsMutable() bool
```
andreyromancev: This method is not necessary. Mutable is IncomingRequest property.

ledger/light/proc/set_result.go
SergeyRadist: What about checkRequestCanChangeState ?

insolar/record/record.go
SergeyRadist: reverted

PR #2826, opened 2019-09-23
INS-3571: remove unneeded code

bigbes:
everything runs under lock, so atomics are unnecessary


logicrunner/executionbroker.go
```diff
@@ -304,7 +308,7 @@ func (q *ExecutionBroker) startProcessors(ctx context.Context) {
 
 func (q *ExecutionBroker) startProcessorCount(ctx context.Context, state *processorState) int {
 	queueLength := state.queue.Length()
-	workerAvailable := state.limiter.Available(ctx)
+	workerAvailable := state.workersLimit - state.workers
```
sexyhamster: I would like this and state.workersLimit == state.workers to be helpers on state, imho it would look better

bigbes:
I was forced to change Reference with RecordReference (for request references) due to some test fails (with separate commit)

logicrunner/executionbroker.go
bigbes: fixed

PR #2827, opened 2019-09-23
NOISSUE: make virtual integration test to support go1.13

bigbes:
Before that patch result of running virtual integration test was:
```
[ 09/23/19 | 14:09:20 ] $ go test -timeout 1200s -tags slowtest ./virtual/integration
flag provided but not defined: -test.testlogfile
Usage of /var/folders/9g/lwhn2csx7rn66b_rg_905wq00000gn/T/go-build606332186/b001/integration.test:
  -verbose-wm
        flag to enable watermill logging
FAIL    github.com/insolar/insolar/virtual/integration  0.021s
FAIL
```

Bug was re order of loading modules:
1) virtual functest (added `verbose-wm` and running `Parse` function), error was here due to absence of `test.testlogfile` command line argument
2) test.testlogfile and other arguments were added later in testing module

### DIFF
--- a/virtual/integration/main_test.go
+++ b/virtual/integration/main_test.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"os"
 	"sync"
 	"testing"
 	"time"
@@ -118,7 +119,6 @@ var verboseWM bool
 
 func init() {
 	flag.BoolVar(&verboseWM, "verbose-wm", false, "flag to enable watermill logging")
-	flag.Parse()
 }
 
 func NewServer(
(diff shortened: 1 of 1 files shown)


virtual/integration/main_test.go
```diff
@@ -197,6 +196,9 @@ func NewServer(
 
 	wmLogger := watermill.LoggerAdapter(watermill.NopLogger{})
 
+	if !flag.Parsed() {
```
sexyhamster: maybe move this into MainTest (I guess it should be there)

bigbes: you're right, https://golang.org/pkg/testing/#hdr-Main describes that

bigbes: fixed

PR #2832, opened 2019-09-23
NOISSUE: Change ID string method to use base64.

andreyromancev:
Change ID string method to use base64.

### DIFF
--- a/reference/local.go
+++ b/reference/local.go
@@ -174,7 +174,7 @@ func (p *byteWriter) isFull() bool {
 
 // String implements stringer on ID and returns base58 encoded value
 func (v Local) String() string {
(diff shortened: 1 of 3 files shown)


reference/local.go
```diff
@@ -174,7 +174,7 @@ func (p *byteWriter) isFull() bool {
 
 // String implements stringer on ID and returns base58 encoded value
 func (v Local) String() string {
-	repr, err := DefaultEncoder().EncodeRecord(&v)
```
bigbes: everytime string method is called new base64 encoder is created? bad idea
BTW, all base64 record identifiers should be prefixed with schema (`EncodingSchema` flag or `FormatSchema`)

PR #2835, opened 2019-09-24
Ins 3585 more restrictions for requests

aMFeriuSS:
**- What I did**
- Added check for oldest opened mutable request. Now we should return an error if the current result trying to close non-oldest opened mutable request.


insolar/payload/error_code.go
```diff
@@ -25,12 +25,13 @@ const (
...
-	CodeReasonIsWrong     = 1009
-	CodeNonClosedOutgoing = 1010
-	CodeNonActivated      = 1011
+	CodeRequestNotFound         = 1006
+	CodeRequestInvalid          = 1007
+	CodeReasonNotFound          = 1008
+	CodeReasonIsWrong           = 1009
+	CodeNonClosedOutgoing       = 1010
+	CodeNonActivated            = 1011
+	CodeNonOldestMutableRequest = 1012
```
SergeyRadist: Group errors please, like
CodeRequestNotFound         = 1006
CodeRequestInvalid    = 1007		
CodeRequestInvalid          = 1007
CodeRequestNonOldestMutable = 1012

aMFeriuSS: done

PR #2837, opened 2019-09-24
Ins 3550 bench check balance from file

megge-dream:
*- What I did**
Add new flag to benchmark to check balances from file before start test.

**- How to verify it**
Run benchmark with --check-every-member


cmd/benchmark/main.go
```diff
@@ -204,14 +206,18 @@ func createMembers(insSDK *sdk.SDK, count int, migration bool) []sdk.Member {
...
 
-func getTotalBalance(insSDK *sdk.SDK, members []sdk.Member) (totalBalance *big.Int) {
+func getTotalBalance(insSDK *sdk.SDK, members []sdk.Member) (*big.Int, map[string]*big.Int) {
 	type Result struct {
 		num     int
 		balance *big.Int
 		err     error
 	}
-
+	totalBalance := &big.Int{}
```
golangcibot: SA4006: this value of `totalBalance` is never used (from `staticcheck`)

cmd/benchmark/main.go
```diff
@@ -253,11 +265,11 @@ func getTotalBalance(insSDK *sdk.SDK, members []sdk.Member) (totalBalance *big.I
 		totalBalance.Add(b, res.balance)
 	}
 
-	return totalBalance
+	return totalBalance, membersWithBalanceMap
 }
 
 func getMembers(insSDK *sdk.SDK, number int, migration bool) ([]sdk.Member, error) {
-	var members = make([]sdk.Member, number)
+	var members = make([]sdk.Member, number+1)
```
afiskon: Where did +1 came from?

megge-dream: We load from file not just number of members, but also migration admin or fee member. Add comment and rewrite it for loadMembers branch only.

PR #2883, opened 2019-09-27
Ins 3631 loop detection

andreyromancev:
Ins 3631 loop detection
**- What I did**

**- How I did it**

**- How to verify it**


insolar/record/record.go
```diff
@@ -270,3 +273,23 @@ func CalculateRequestAffinityRef(
 	}
 	return affinityRef
 }
+
+// ObjectIDFromRequest calculates object is from request. Passed hasher must be newly created.
+func ObjectIDFromRequest(hasher hash.Hash, request Request, requestID insolar.ID) (insolar.ID, error) {
```
sexyhamster: if you need to have fresh hasher, maybe it is better to get ReferenceHasher, that will allow you to create hasher within the function?

ledger/light/executor/request_checker.go
```diff
@@ -355,3 +348,19 @@ func (c *RequestCheckerDefault) getRequestLocal(
...
 }
+
+func hasAPIRequest(reqs []record.CompositeFilamentRecord, apiRequest string) bool {
+	for _, req := range reqs {
+		switch r := record.Unwrap(&req.Record.Virtual).(type) {
+		case *record.IncomingRequest:
+			if r.APIRequestID == apiRequest {
+				return true
+			}
+		case *record.OutgoingRequest:
```
ruz: I believe you shouldn't check outgoing requests.

PR #2884, opened 2019-09-27
NOISSUE: always rebuild backupmanager during slow_tests

afiskon:
Note: no idea what BIN_DIR was used for.

### DIFF
--- a/cmd/backupmanager/backupmanager_test.go
+++ b/cmd/backupmanager/backupmanager_test.go
@@ -38,16 +38,19 @@ import (
 var binaryPath string
 
 func init() {
-	var ok bool
+	wd, err := os.Getwd()
+	binaryPath = filepath.Join(wd, "..", "..", "bin")
 
-	binaryPath, ok = os.LookupEnv("BIN_DIR")
-	if !ok {
-		wd, err := os.Getwd()
-		binaryPath = filepath.Join(wd, "..", "..", "bin")
+	if err != nil {
+		panic(err.Error())
+	}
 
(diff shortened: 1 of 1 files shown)


cmd/backupmanager/backupmanager_test.go
```diff
@@ -38,16 +38,19 @@ import (
...
-		binaryPath = filepath.Join(wd, "..", "..", "bin")
+	if err != nil {
+		panic(err.Error())
+	}
 
-		if err != nil {
-			panic(err.Error())
-		}
+	// Always rebuild backupmanager
+	bashCmd := "cd " + binaryPath + " && (rm backupmanager || true) && go build ../cmd/backupmanager"
```
dmitry-panchenko: You could use `rm -f` to avoid ugly `|| true`

afiskon: Nope:
```
$ rm -r nosuchfile
rm: nosuchfile: No such file or directory
$ echo $?
1
```

afiskon: I believe you are right, tnx.

afiskon: What I personally dislike about -f is that it may remove a write-protected file.

dmitry-panchenko: Yeah, but `backupmanager ` could be write-protected?
Your point is fine when you are trying to delete a user input data but in this case you we could ignore "write-protected" part.

PR #2885, opened 2019-09-27
Ins 3624 availability tests

LinnikD:
added availability to functest launcher

added gracefull stop for future testing

fixed issue with not working response parcing in checker


api/availability.go
```diff
@@ -67,13 +69,26 @@ func (nc *NetworkChecker) Start(ctx context.Context) error {
...
+			}
 		}
+		nc.stopped <- struct{}{}
 	}(ctx)
 	return nil
 }
 
+func (nc *NetworkChecker) Stop() {
+	nc.stopped <- struct{}{}
+	<-nc.stopped
```
afiskon: FYI: Since you are using a non-buffered channel you can avoid making a double check here. When first `<-` unblocks you can be sure that the goroutine is about to terminate.

PR #2887, opened 2019-09-27
INS-3628_outgoing_deduplication no more double call of outgoing

aalekseev-ru:
**- What I did**
1) remove some useless code
2) add check for already processed outgoing requests
3) add check for Saga's outgoing requests with results (it must never happen)

**- How I did it**
write code and tests
**- How to verify it**
run tests

### DIFF
--- a/logicrunner/rpc_methods.go
+++ b/logicrunner/rpc_methods.go
@@ -189,18 +189,21 @@ func (m *executionProxyImplementation) RouteCall(
 	if req.Saga {
 		// Saga methods are not executed right away. LME will send a method
 		// to the VE when current object finishes the execution and validation.
+		if outReqInfo.Result != nil {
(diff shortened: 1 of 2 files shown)


logicrunner/rpc_methods.go
```diff
@@ -189,18 +189,21 @@ func (m *executionProxyImplementation) RouteCall(
 	if req.Saga {
 		// Saga methods are not executed right away. LME will send a method
 		// to the VE when current object finishes the execution and validation.
+		if outReqInfo.Result != nil {
```
sexyhamster: why is it an error? maybe something was broken in between or discarded by heavy

afiskon: If we are processing a request that creates a saga and saga is already closed something is VERY wrong with the system. There is no valid scenario when it can happen. Personally I suggested to `panic()` in this case.

PR #2897, opened 2019-09-28
add test that passes a coin from account to another in parallel

ruz:
pass one coin in parallel, make sure balance settles

### DIFF
--- a/functest/pressure_test.go
+++ b/functest/pressure_test.go
@@ -20,10 +20,13 @@ package functest
 
 import (
 	"sync"
+	"sync/atomic"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 
+	"github.com/insolar/insolar/insolar"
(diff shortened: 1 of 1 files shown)


functest/pressure_test.go
```diff
@@ -135,3 +138,113 @@ func (c *One) Dec() (int, error) {
...
+		require.Greater(syncT, successes, uint32(0))
+		require.Greater(syncT, errors, uint32(0))
+
+		r1 := callMethod(syncT, w1Ref, "Balance")
+		require.Empty(syncT, r1.Error)
+		r2 := callMethod(syncT, w2Ref, "Balance")
+		require.Empty(syncT, r2.Error)
+
+		pass := false
+		for i := 0; i < 10; i++ {
```
aalekseev-ru: why do we need a loop here if we have waitgroup?

PR #2921, opened 2019-09-30
INS-3652: add new obj record func and use it in contract

megge-dream:
**- What I did**
Add NewObjectReferenceFromBase58 func and use it in contract

**- How to verify it**
Run tests

### DIFF
--- a/insolar/record.go
+++ b/insolar/record.go
@@ -18,10 +18,10 @@ package insolar
 
 import (
 	"encoding/binary"
-	"fmt"
 
 	"github.com/insolar/insolar/longbits"
 	"github.com/insolar/insolar/reference"
+	"github.com/pkg/errors"
 )
 
 const (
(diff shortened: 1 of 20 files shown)


insolar/record.go
```diff
@@ -59,6 +60,18 @@ func NewGlobalReference(local ID, base ID) *Reference {
...
 }
 
+// NewObjectReferenceFromBase58 deserializes reference from base58 encoded string and checks if it object reference
+func NewObjectReferenceFromBase58(input string) (*Reference, error) {
+	global, err := NewReferenceFromBase58(input)
+	if err != nil {
+		return nil, err
+	}
+	if !global.IsObjectReference() {
+		return nil, fmt.Errorf("provided reference is not object")
```
ita-sammann: errors.New() is better here, you don't need formatting

megge-dream: Done!

PR #2946, opened 2019-10-01
don't flush ledgerHasMoreRequests when we had incomming during fetch

ruz:
fetch at least one more time, it's possible that there is a logic
race between fetch, insert into ledger and setting flag.

If channel is not empty then we will start another fetch right away,
so it's sort of double commit

### DIFF
--- a/logicrunner/executionbroker.go
+++ b/logicrunner/executionbroker.go
@@ -151,6 +151,7 @@ func (q *ExecutionBroker) noMoreRequestsOnLedger(ctx context.Context) {
 	defer q.stateLock.Unlock()
 
 	if len(q.probablyMoreSinceLastFetch) == 0 {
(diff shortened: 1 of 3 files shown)


logicrunner/executionbroker.go
```diff
@@ -147,11 +146,13 @@ func (q *ExecutionBroker) AbandonedRequestsOnLedger(ctx context.Context) {
 	go q.startProcessor(ctx)
 }
 
-func (q *ExecutionBroker) NoMoreRequestsOnLedger(ctx context.Context) {
+func (q *ExecutionBroker) noMoreRequestsOnLedger(ctx context.Context) {
```
golangcibot: `(*ExecutionBroker).noMoreRequestsOnLedger` - `ctx` is unused (from `unparam`)

PR #2948, opened 2019-10-02
INS-3633: fix in executionbroker tests leaking goroutines

sexyhamster:
**- What I did**
fixed leaking goroutines in executionbroker tests

**- How I did it**
with leaktest

**- How to verify it**
run tests

### DIFF
--- a/logicrunner/executionbroker.go
+++ b/logicrunner/executionbroker.go
@@ -281,7 +281,7 @@ func (q *ExecutionBroker) processTranscript(ctx context.Context, transcript *com
 		q.finishTranscript(ctx, transcript)
 
 		if sendReply {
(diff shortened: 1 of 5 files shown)


logicrunner/executionbroker.go
```diff
@@ -412,7 +412,6 @@ func (q *ExecutionBroker) OnPulse(ctx context.Context) []payload.Payload {
 		// clean everything, just in case
 		q.finished = nil
 		q.deduplicationTable = make(map[insolar.Reference]bool)
-		q.pending = insolar.InPending
 		q.PendingConfirmed = true
```
sexyhamster: @bigbes need your review for this change. I'm pretty sure now there is no need to to it

sexyhamster: disregard that

PR #2958, opened 2019-10-03
additional logs for search_index + fixes for deduplication

egorikas:
**- What I did**

**- How I did it**

**- How to verify it**


ledger/heavy/proc/search_index.go
```diff
@@ -68,6 +69,12 @@ func (p *SearchIndex) Proceed(ctx context.Context) error {
 		return errors.Wrap(err, "failed to unmarshal searchIndex message")
 	}
 
+	logger := inslogger.FromContext(ctx).WithFields(map[string]interface{}{
+		"objID": searchIndex.ObjectID.DebugString(),
```
ruz: use `object_id` like we do in other cases

PR #2975, opened 2019-10-07
Make sure that the flushing procedure can't be interfered with wrappi…

cyraxred:
Log output may be truncated on fatal.
This fix changes logger access to the output to ensure that the flushing procedure can't be "fenced off" by wrappers of io.Writer.


log/logoutput/outpur_adapter.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	if w, ok := output.(insolar.LogLevelWriter); ok {
+		return &Adapter{output: w, flushFn: flushFn, state: uint32(flags)}
+	}
+
+	return &Adapter{output: writerAdapter{output}, flushFn: flushFn, fatalFlushFn: fatalFlushFn, state: uint32(flags)}
+}
+
+var ClosedError = errors.New("closed")
```
golangcibot: error var ClosedError should have name of the form ErrFoo (from `golint`)

PR #3004, opened 2019-10-17
Ins 3735 new backup scheme

ivanshibitov:
**- What I did**
1. Drop old logic with background running of badger's value GC
2. Run GC only after pulse finalization with given( in config ) frequency

**- How I did it**

**- How to verify it**

### DIFF
--- a/ledger/heavy/executor/finalize_pulse.go
+++ b/ledger/heavy/executor/finalize_pulse.go
@@ -27,6 +27,33 @@ import (
 	"github.com/insolar/insolar/ledger/object"
 )
 
+type BadgerGCRunner interface {
+	// RunValueGC run badger values garbage collection
+	RunValueGC(ctx context.Context)
+}
+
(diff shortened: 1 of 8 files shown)


ledger/heavy/executor/finalize_pulse.go
```diff
@@ -27,6 +27,33 @@ import (
...
+	return &BadgerGCRunInfo{
+		runner:       runner,
+		runFrequency: runFrequency,
+	}
+}
+
+func (b *BadgerGCRunInfo) RunGCIfNeeded(ctx context.Context) {
+	b.callCounter++
+	if (b.runFrequency > 0) && (b.callCounter >= b.runFrequency) && (b.callCounter%b.runFrequency == 0) {
+		b.runner.RunValueGC(ctx)
```
afiskon: callCounter should be set to zero, otherwise GC will be called only once

PR #3026, opened 2019-10-25
INS-3745: temp fix of leaktest issue

Vany:
INS-3745: temp fix of leaktest issue


ita-sammann:
WTF is this?!

Vany:
> NOOOOOOOOOOOOOOOOOOOO

it's too bore =)

logicrunner/writecontroller/writecontroller_test.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 func TestWriteController_Open(t *testing.T) {
-	t.Parallel()
```
sexyhamster: This should be disabled or it is forgotten

logicrunner/logicrunner_unit_test.go
```diff
@@ -59,6 +59,15 @@ import (
 
 var _ insolar.LogicRunner = &LogicRunner{}
 
+const useLeakTest = true
```
sexyhamster: do we still need it? I'd rather have it on everywhere where it can be on. Except parallel, where it could be optional

sexyhamster: I mean `func leakTestCheck`

Vany: sec pls

Vany: like this

PR #3030, opened 2019-10-29
ins ref check for record

LinnikD:
ins ref check for record


insolar/record.go
```diff
@@ -75,6 +75,18 @@ func NewObjectReferenceFromString(input string) (*Reference, error) {
 	return global, nil
 }
 
+// NewRecordReferenceFromString deserializes reference from base64 encoded string and checks if it record reference
```
ita-sammann: ... if it _is_ record reference :)

PR #3035, opened 2019-11-05
remove VestingType from Deposit contract, it's incomplete and not required

ruz:
this functionality was never complete and obsoleted by Vesting and
VestingStep fields


application/builtin/contract/deposit/deposit.go
```diff
@@ -280,7 +276,7 @@ func (d *Deposit) canTransfer(transferAmount *big.Int) error {
 			c++
 		}
 	}
-	if d.VestingType == appfoundation.DefaultVesting && c < numConfirmation {
+	if c < numConfirmation {
```
megge-dream: With this change all deposits from Genesis will be block forever, cause they don't have confirms by migration daemons.

ruz: that's bad, will fix

PR #3044, opened 2019-11-08
NOISSUE: temporary disable goproxy until goproxy.insolar.io will be up

afiskon:
temporary disable goproxy


Makefile
```diff
@@ -1,5 +1,4 @@
 export GO111MODULE ?= on
-export GOPROXY ?= https://goproxy.io
 export GOSUMDB ?= sum.golang.org
```
AndreyBronin: just note - GOSUMDB is not supported in 1.12

PR #3048, opened 2019-11-10
NOISSUE: try to optimize send message

krestkrest:
**- What I did**
Increased performance of watermill message sending a little bit. I got 10 -> 14 resp/s increase for benchmark on my lame notebook. Benchmark configuration: 5 discovery nodes, info log level, bench -c 10 -r 100.
**- How I did it**
1. Removed extra serialization/deserialization calls;
2. Replaced gob with protobuf serialization.

**- How to verify it**
Run unit & integration tests and benchmark


insolar/bus/bus.go
```diff
@@ -469,7 +469,7 @@ func (b *Bus) wrapMeta(
 ) (payload.Meta, *message.Message, error) {
 	msg = msg.Copy()
 
-	meta := payload.Meta{
+	payloadMeta := payload.Meta{
 		Polymorph:  uint32(payload.TypeMeta),
 		Payload:    msg.Payload,
 		Receiver:   receiver,
```
krestkrest: `Receiver` is no longer used anywhere except tests, but I did not dare to remove it from payload.Meta protobuf

network/servicenetwork/serialize.go
```diff
@@ -52,39 +52,88 @@ package servicenetwork
...
+		return "", errors.Wrap(err, "failed to read string data")
+	}
+	return string(result), nil
+}
+
+func serializeMetadata(buffer *bytes.Buffer, metadata message.Metadata) {
+	keyValuePairsCount := bits.NewVarSize(len(metadata))
+	_, _ = buffer.Write(keyValuePairsCount.Serialize())
+
+	for key, value := range metadata {
```
krestkrest: The key iteration order in map is guaranteed to be randomized by Golang spec, so serialized representation will not be stable. I thought of sorting the keys before serialization, however, gob (that was used previously) does not sort map keys either. So I decided to keep that behaviour unchanged for now.

network/servicenetwork/serialize.go
```diff
@@ -52,39 +52,88 @@ package servicenetwork
...
+
+	"github.com/insolar/insolar/insolar/bits"
 )
 
-// serializeMessage returns io.Reader on buffer with encoded message.Message (from watermill).
-func serializeMessage(msg *message.Message) (io.Reader, error) {
+func serializeString(buffer *bytes.Buffer, s string) {
+	// here and below: bytes.Buffer never returns errors, so do not check them
+	size := bits.NewVarSize(len(s))
+	_, _ = buffer.Write(size.Serialize())
```
sexyhamster: please do not skip error checks

network/servicenetwork/serialize.go
```diff
@@ -52,39 +52,88 @@ package servicenetwork
...
 
 	"github.com/ThreeDotsLabs/watermill/message"
 	"github.com/pkg/errors"
+
+	"github.com/insolar/insolar/insolar/bits"
 )
 
-// serializeMessage returns io.Reader on buffer with encoded message.Message (from watermill).
-func serializeMessage(msg *message.Message) (io.Reader, error) {
+func serializeString(buffer *bytes.Buffer, s string) {
```
sexyhamster: i wonder why you didn't check msgpack serialization instead of homemade, maybe add benchmarks?

network/servicenetwork/watermill.go
```diff
@@ -88,17 +88,15 @@ func (n *ServiceNetwork) SendMessageHandler(msg *message.Message) error {
...
-	err := meta.Unmarshal(msg.Payload)
-	if err != nil {
-		return errors.Wrap(err, "failed to unwrap message")
+	receiver := msg.Metadata.Get(meta.Receiver)
+	if receiver == "" {
+		return errors.New("failed to send message: Receiver in message metadata is not set")
 	}
-	if meta.Receiver.IsEmpty() {
-		return errors.New("failed to send message: Receiver in meta message not set")
+	node, err := insolar.NewReferenceFromString(receiver)
```
sexyhamster: probably you still need here `node.Receiver.IsEmpty()`

network/servicenetwork/serialize.go
krestkrest: msgpack does not guarantee map keys order in serialized message: https://github.com/msgpack/msgpack/issues/215 
The question is: are there any plans to include message in hash/signature calculation? If no, then there is no need for custom serialization, msgpack is x2 faster. If yes, then there is an option to make keys sorted in custom serialization by adding 3 additional lines of code - and the serialization will be stable (in contrast to msgpack)

sexyhamster: from my point of view, there is no point of stable serialization for metadata. So it's better just to use protobuf, thanks for idea https://github.com/insolar/insolar/pull/3058

insolar/bus/bus.go
```diff
@@ -479,13 +479,14 @@ func (b *Bus) wrapMeta(
 		ID:         []byte(msg.UUID),
 	}
 
-	buf, err := meta.Marshal()
+	buf, err := payloadMeta.Marshal()
 	if err != nil {
 		return payload.Meta{}, nil, errors.Wrap(err, "wrapMeta. failed to wrap message")
 	}
 	msg.Payload = buf
+	msg.Metadata.Set(meta.Receiver, receiver.String())
```
sexyhamster: @megge-dream for me this one actually looks good, I wonder if we can switch to medata field?

network/servicenetwork/watermill.go
krestkrest: yeap, fixed

network/servicenetwork/serialize.go
krestkrest: replaced with protobuf

dmitry-panchenko:
I'm going to merge this after tests are passed #3060

krestkrest:
`TestMessageBus_SendTarget` looks a little bit senseless since both Payload and Metadata of the message are mutated in `wrapMeta`

dmitry-panchenko:
> `TestMessageBus_SendTarget` looks a little bit senseless since both Payload and Metadata of the message are mutated in `wrapMeta`

PR #3058, opened 2019-11-13
switched network message transport to protobuf

sexyhamster:
**- What I did**
switched message serialization from gob to protobuf


sexyhamster:
closing in favor for #3048

PR #3068, opened 2019-11-18
NOISSUE: Makefile refactoring

Tsovak:
**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -241,7 +241,7 @@ ci_test_func:  ## run functest 3 times
 		$(MAKE) ci-test-func-base
 
 .PHONY: ci-test-func-long
-ci-test-func-long: ## run functest with race
+ci-test-func-long: ## run functest with race and a little count
 	CI_TEST_ARGS=" -p 10 -race " \
 	FUNCTEST_COUNT=10 \
 		$(MAKE) ci-test-func-base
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -24,7 +24,7 @@ MOCKS_PACKAGE = github.com/insolar/insolar/testutils
 GOBUILD ?= go build -mod=vendor
 GOTEST ?= go test -mod=vendor
 
-FUNCTEST_COUNT ?= 1
+FUNCTEST_COUNT ?= 3
```
afiskon: I'm afraid you can't do it. This variable is used in `make functest`. Developers use it locally as well. Please re-define environment variable on CI if necessary.

Makefile
```diff
@@ -208,24 +208,24 @@ ci_test_with_coverage: ## run unit tests with coverage, outputs json to stdout (
...
-		$(GOTEST) $(CI_TEST_ARGS) $(TEST_ARGS) -json -v -failfast -tags slowtest ./... -count 1 | tee -a ci_test_unit.json
+		$(GOTEST) $(CI_TEST_ARGS) $(TEST_ARGS) -json -v -failfast -tags slowtest ./... -count 1
 
 .PHONY: ci_test_func
 ci_test_func: ## run functest 3 times, redirects json output to file (CI)
 	# GOMAXPROCS=2, because we launch at least 5 insolard nodes in functest + 1 pulsar,
 	# so try to be more honest with processors allocation.
 	GOMAXPROCS=$(CI_GOMAXPROCS) CGO_ENABLED=1  \
-		$(GOTEST) $(CI_TEST_ARGS) $(TEST_ARGS) -json -tags "functest bloattest" -v ./application/functest -count 3 -failfast | tee ci_test_func.json
+		$(GOTEST) $(CI_TEST_ARGS) $(TEST_ARGS) -json -tags "functest bloattest" -v ./application/functest -count=$(FUNCTEST_COUNT) -failfast
```
ita-sammann: CI functests should run 3 iterations

Tsovak: you right. I pushed the new changes

afiskon: This code is quite pointless. Since it doesn't allow to override FUNCTEST_COUNT from outside the script you could just use `-count 3` with the exact same effect.

ita-sammann: Why do you redefine the variable? It's very unobvious. Just hardcode "3"

PR #3072, opened 2019-11-18
NOISSUE: improve build tooling (PoC)

nordicdyno:
* move binary tools dependencies to separate module hidden for go tools `build/_go_tools`
* add `vendor/` directory check script


nordicdyno:
It's solves issue with Go 1.13 (visible dependencies in `tools.go` for go tools), so till we stuck on Go 1.12 branch - won't fix.

PR #3086, opened 2019-11-25
INS-3925: use the same lightchainlimit as @ prod in launchnet.sh

afiskon:
This suppose to fix `last finalized pulse falls behind too much. Stop node` during nightly functests run.


ruz:
I think @andreyromancev should make decision on this. Looks like workaround. I don't think that nightly functests generate enough for finalization to fallback 50 seconds behind.

PR #3088, opened 2019-11-27
NOISSUE: deposit tweaks

ita-sammann:
* Added restriction to deposit constructor that only allows creating a deposit if vesting period is a multiple of vesting step. To avoid creating deposits with "fractional" vesting steps.
 * Refactored all errors handling in the deposit (`errors.Wrap()` instead of `fmt.Errorf`).
 * Rewritten `canTransfer()` method in a clearer way.

### DIFF
--- a/application/builtin/contract/deposit/deposit.go
+++ b/application/builtin/contract/deposit/deposit.go
@@ -21,6 +21,8 @@ import (
 	"math/big"
 	"strings"
 
+	"github.com/pkg/errors"
+
 	"github.com/insolar/insolar/application/appfoundation"
 	"github.com/insolar/insolar/application/builtin/proxy/deposit"
 	"github.com/insolar/insolar/application/builtin/proxy/member"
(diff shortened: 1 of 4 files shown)


application/builtin/contract/deposit/deposit.go
```diff
@@ -294,48 +300,65 @@ func (d *Deposit) checkConfirm(migrationDaemonRef string, amountStr string) (boo
...
+	passedSteps := big.NewInt(int64(currentPulse-d.PulseDepositUnHold) / d.VestingStep)
+	// Amount that has been vested by now
+	vestedByNow := new(big.Int).Div(
+		new(big.Int).Mul(amount, passedSteps),
+		totalSteps,
 	)
+	// Amount that is still locked on deposit
+	onHold := new(big.Int).Sub(amount, vestedByNow)
+	// Amount that is now available for withdrawal
+	availableNow := new(big.Int).Sub(balance, onHold)
```
ruz: this can become negative when balance is 0 and vesting already started

ita-sammann: Fixed. Returning zero in this case

application/builtin/contract/deposit/deposit.go
```diff
@@ -52,6 +54,10 @@ type Deposit struct {
 // New creates new deposit.
 func New(txHash string, lockup int64, vesting int64, vestingStep int64) (*Deposit, error) {
 
+	if vesting%vestingStep != 0 {
+		return nil, errors.New("vesting is not multiple of vestingStep")
```
LinnikD: this check is also needed in bootstrap, when vesting and vestingStep is configured

ita-sammann: Fixed

PR #3097, opened 2019-12-02
INS-3925: fix off-by-one error

afiskon:
See comments.


ledger/heavy/executor/finalization_keeper.go
```diff
@@ -73,7 +73,9 @@ func (f *FinalizationKeeperDefault) OnPulse(ctx context.Context, current insolar
 		return errors.New(fmt.Sprintf("Current pulse ( %d ) is less than last confirmed ( %d )", current, lastConfirmedPulse))
 	}
 
-	if lastConfirmedPulse <= bottomLevel.PulseNumber {
+	// Note: We can't do '<=' here. Consider the case when we just started and there is
+	// only one finalized pulse. In this case bottomLevel == lastConfirmedPulse and it's OK.
+	if lastConfirmedPulse < bottomLevel.PulseNumber {
```
ruz: this analysis is incorrect. pulseCalculator returns ErrNotFound in described case. We don't get to this point until DB has enough pulses.

ivanshibitov: and we even have such check

PR #3105, opened 2019-12-04
INS-3655: recover panics in contracts

bigbes:
* also make panics in contract execution as user error (store it in DB
  as part of response)

### DIFF
--- a/logicrunner/builtin/builtin.go
+++ b/logicrunner/builtin/builtin.go
@@ -74,15 +74,21 @@ func NewBuiltIn(am artifacts.Client, stub LogicRunnerRPCStub) *BuiltIn {
 }
 
 func (b *BuiltIn) CallConstructor(
-	ctx context.Context, callCtx *insolar.LogicCallContext, codeRef insolar.Reference,
-	name string, args insolar.Arguments,
-) ([]byte, insolar.Arguments, error) {
+	ctx context.Context,
+	callCtx *insolar.LogicCallContext,
+	codeRef insolar.Reference,
+	name string,
+	args insolar.Arguments,
+) (
+	[]byte, insolar.Arguments, error,
+) {
 	executeStart := time.Now()
 	ctx = insmetrics.InsertTag(ctx, metrics.TagContractPrototype, b.PrototypeRefRegistry[codeRef])
 	ctx = insmetrics.InsertTag(ctx, metrics.TagContractMethodName, "Constructor")
-	defer func(ctx context.Context) {
 
-		stats.Record(ctx, metrics.ContractExecutionTime.M(float64(time.Since(executeStart).Nanoseconds())/1e6))
+	defer func(ctx context.Context) {
+		executionTime := time.Since(executeStart).Nanoseconds()
+		stats.Record(ctx, metrics.ContractExecutionTime.M(float64(executionTime)/1e6))
 	}(ctx)
 
 	ctx, span := instracer.StartSpan(ctx, "builtin.CallConstructor")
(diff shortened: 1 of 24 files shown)


logicrunner/preprocessor/templates/wrapper.go.tpl
```diff
@@ -94,40 +96,69 @@ func INSMETHOD_GetPrototype(object []byte, data []byte) ([]byte, []byte, error)
...
+		)
+	}
+
+	needRecover := true
+	defer func() {
+		if !needRecover {
+			return
+		}
+		if r := recover(); r != nil {
+			recoveredError := errors.Wrap(errors.Errorf("%v", r), "Failed to execute method (panic)")
```
ruz: can you move this repeated code into a function?

bigbes: i cannot, it contains templated part :(

logicrunner/preprocessor/main_test.go
```diff
@@ -597,47 +597,6 @@ func ( A ) Get() error {
 	s.NotContains(bufWrapper.String(), `"some/test/import/path"`)
 }
 
-func (s *PreprocessorSuite) TestImportsFromContractUseForReturnValue() {
```
ruz: I don't understand why you deleted this test. Can you explain?

bigbes: We're importing packages needed for declaration of contract return values, now. That test tests that we won't import that packages.

ruz: Nope, that test was checking that we deal correctly with return values of type declared in an external package. This is what it was testing. If you changed the way it works then you should change the test. Don't delete it.

logicrunner/preprocessor/templates/wrapper.go.tpl
```diff
@@ -94,40 +96,69 @@ func INSMETHOD_GetPrototype(object []byte, data []byte) ([]byte, []byte, error)
...
+			recoveredError = ph.MakeErrorSerializable(recoveredError)
+
+			if PanicIsLogicalError {
+				ret{{ $method.LastErrorInRes }} = recoveredError
+
+				newState = object
+				err = serializeResults()
```
ruz: this part doesn't look correct, it should be:
```
err = serializeResults()
if err != nil {
    newState = object
}
```

logicrunner/builtin/builtin.go
```diff
@@ -106,13 +112,23 @@ func (b *BuiltIn) CallConstructor(
...
+	data []byte,
+	method string,
+	args insolar.Arguments,
+) (
+	[]byte, insolar.Arguments, error,
+) {
 	executeStart := time.Now()
 	ctx = insmetrics.InsertTag(ctx, metrics.TagContractPrototype, b.PrototypeRefRegistry[codeRef])
-	ctx = insmetrics.InsertTag(ctx, metrics.TagContractMethodName, method)
+	ctx = insmetrics.InsertTag(ctx, metrics.TagContractMethodName, "Constructor")
```
ruz: You may have multiple constructors with different names. I don't mind a prefix or something else added here.  Please split out these changes to metrics from this PR. They have nothing to do with recovering panics in contracts.

bigbes: fixed

logicrunner/preprocessor/templates/wrapper.go.tpl
bigbes: fixed

logicrunner/preprocessor/main_test.go
bigbes: fixed

PR #3108, opened 2019-12-05
INS-3931: we think that it's more correct to have this funds on account

ruz:
no deposits for these 5, directly on accounts.


application/functest/fund_test.go
```diff
@@ -91,54 +91,6 @@ func checkBalanceAndDepositFewTimes(t *testing.T, m *launchnet.User, expectedBal
 		depositStr, expectedDeposit)
 }
 
-func TestFundsTransferDeposit(t *testing.T) {
```
LinnikD: add check balance tests on new accounts

ruz: we have already this test and I adjust it. Take closer look.

PR #3125, opened 2019-12-12
swagger spec restrict and benchmark migration test

sexyhamster:
**- What I did**
- fix in benchmark scenario to work according to specification


sexyhamster:
closing in favor #3128 3128

PR #3147, opened 2019-12-17
INS-3961: context.Canceled test

ElSacramento:
**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/ledger/heavy/exporter/record_exporter_server_test.go
+++ b/ledger/heavy/exporter/record_exporter_server_test.go
@@ -585,6 +585,18 @@ func TestRecordServer_Export_Composite(t *testing.T) {
 		require.Equal(t, secondRec, resRecord.Record)
 	})
 
+	t.Run("context.Canceled error", func(t *testing.T) {
+		streamMock := &streamMock{checker: func(i *Record) error {
+			return context.Canceled
+		}}
+
+		err := recordServer.Export(&GetRecords{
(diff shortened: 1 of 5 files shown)


ledger/heavy/exporter/record_exporter_server_test.go
```diff
@@ -585,6 +585,18 @@ func TestRecordServer_Export_Composite(t *testing.T) {
 		require.Equal(t, secondRec, resRecord.Record)
 	})
 
+	t.Run("context.Canceled error", func(t *testing.T) {
+		streamMock := &streamMock{checker: func(i *Record) error {
```
andreyromancev: Shadowing the type name. Let's assign a different name for it.

ledger/heavy/exporter/record_exporter_server_test.go
```diff
@@ -585,6 +585,18 @@ func TestRecordServer_Export_Composite(t *testing.T) {
...
+			return context.Canceled
+		}}
+
+		err := recordServer.Export(&GetRecords{
+			PulseNumber:  firstPN,
+			RecordNumber: 1,
+			Count:        1,
+		}, streamMock)
+		require.Error(t, err)
```
andreyromancev: If we want to test a specific behaviour, we should assert a specific error. This will assert a general error that could be caused by an unexpected behaviour.

PR #3152, opened 2019-12-18
INS-3971 badger start time metric added

SergeyRadist:
**- What I did**
added metric on opening badger
**- How I did it**

**- How to verify it**

### DIFF
--- a/server/internal/heavy/metrics.go
+++ b/server/internal/heavy/metrics.go
@@ -1,0 +1,44 @@
+//
+// Copyright 2019 Insolar Technologies GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


server/internal/heavy/metrics.go
```diff
@@ -0,0 +1,44 @@
...
+
+package heavy
+
+import (
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
+)
+
+var (
+	statBadgerStartTime = stats.Float64(
```
LinnikD: is it first metrics of heavy node?   There is a finalisation delta somwhere in code, why do we need new file?

SergeyRadist: hm, first in heavy package, we can add it to executor package and make it public, but I think it will be less obvious

PR #3165, opened 2020-01-15
NOISSUE: refactor dockerfile

Tsovak:
we forgot to change a build in the dockerfile

**- What I did**
Changed the build, create the run.sh which passes needed args to build 

**- How I did it**

**- How to verify it**

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,7 +1,10 @@
 .artifacts
-data
-testdata
+.github
 bin
+ci
+data
+docker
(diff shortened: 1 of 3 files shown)


docker/run.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
nordicdyno: what is purpose of script? it think it would be nice add explanation here

PR #3183, opened 2020-01-24
INS-4139: fix DBStore.TruncateHead method

afiskon:
Fix off-by one error in dbStore.TruncateHead implementation.


afiskon:
OK we figured out that 1) comments lie 2) Most likely fromPulse should be included, thus all classes except pulseDB are implemented correctly. There will be another PR.

PR #3188, opened 2020-01-27
GH actions release logic

yellowmegaman:
- What I did
Added release workflow to the repo
- How I did it
I've added Dockerfile and GH actions workflow pipeline which is being triggered by tag event.
Every tag event will trigger build, after that docker image will be created and pushed to docker registry and new GH release will be added to releases page, w/ binaries produced by build stage.
- Requirements
Those secrets should be added to the repo for workflow to work properly:
```
REGISTRY_URL
REGISTRY_USERNAME
REGISTRY_PASSWORD
```
- How to verify it
You can tag any commit with any tag name like that:
```
$ git tag -a 'v1.0.0' -m 'v1.0.0' 
$ git push origin v1.0.0
```
After steps above you should be able to download artifacts from release page and pull image from registry:
```
docker pull insolar/insolar:v1.0.0
```
- Description for the changelog
GH Actions release pipeline

### DIFF
--- a/.github/workflows/tag.yaml
+++ b/.github/workflows/tag.yaml
@@ -1,0 +1,81 @@
+name: insolar
+on:
+  push:
+    tags:
+      - '*'
+jobs:
+  build-pack-release:
+    name: build-pack-release
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout code
+        uses: actions/checkout@master
+      - name: gather info
+        id: info
+        run: |
+          echo "::set-output name=tag_name::$(echo $GITHUB_REF | cut -d '/' -f3)"
+          echo "::set-output name=sha_short::$(git rev-parse --short HEAD)"
+          echo "::set-output name=go_version::1.12.15"
+          echo "::set-output name=build_date::$(shell ./scripts/dev/git-date-time.sh -d)"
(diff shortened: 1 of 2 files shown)


.github/workflows/tag.yaml
```diff
@@ -0,0 +1,81 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout code
+        uses: actions/checkout@master
+      - name: gather info
+        id: info
+        run: |
+          echo "::set-output name=tag_name::$(echo $GITHUB_REF | cut -d '/' -f3)"
+          echo "::set-output name=sha_short::$(git rev-parse --short HEAD)"
+          echo "::set-output name=go_version::1.12.15"
```
ita-sammann: The version here doesn't match one in Dockerfile. Can we somehow use something like 1.12.<latest> everywhere?

yellowmegaman: Sure! Updated.

nordicdyno: we can pass go version as argument to dockerfile and use GHA variable to pass it

nordicdyno: oh, I see go version is passes as argument to Dockerfile, so it would be nice just set expected Go version (last 1.12.x patch) in Dockerfile

Tsovak:
the changes don't depend on tests

PR #3227, opened 2020-02-17
NOISSUE: various fixes after Badger + PostgreSQL merge

afiskon:
1) Don't write an "error" that happens on regular basis.
2) Quick fix of nil pointer dereference (will be refactored properly later)

### DIFF
--- a/insolar/jet/postgres_dbstore.go
+++ b/insolar/jet/postgres_dbstore.go
@@ -139,7 +139,8 @@ func (s *PostgresDBStore) get(pn insolar.PulseNumber) *Tree {
 	err = row.Scan(&serializedTree)
 
 	if err != nil {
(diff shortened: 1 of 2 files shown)


afiskon:
@andreyromancev I'm not sure if this is OK but it happens a lot. Should we be concerned?

insolar/jet/postgres_dbstore.go
```diff
@@ -139,7 +139,8 @@ func (s *PostgresDBStore) get(pn insolar.PulseNumber) *Tree {
 	err = row.Scan(&serializedTree)
 
 	if err != nil {
-		log.Errorf("PostgresDBStore.get - row.Scan failed: %v", err)
+		// this happens on regular basis
+		// log.Errorf("PostgresDBStore.get - row.Scan failed: %v", err)
```
andreyromancev: Let's just remove this line.

PR #3229, opened 2020-02-17
Penv 69 kube deploy support

nordicdyno:
add kubernetes deploy support for https://github.com/insolar/kube

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,5 +24,12 @@ WORKDIR /go/src/github.com/insolar/insolar
 RUN  set -eux; \
      groupadd -r insolar --gid=999; \
      useradd -r -g insolar --uid=999 --shell=/bin/bash insolar
-COPY --from=build /go/src/github.com/insolar/insolar/bin/insolar /go/src/github.com/insolar/insolar/bin/insolard /go/src/github.com/insolar/insolar/bin/keeperd /go/src/github.com/insolar/insolar/bin/pulsard /usr/local/bin/
-
+COPY --from=build /go/src/github.com/insolar/insolar/application/api/spec/api-exported.yaml /app/api-exported.yaml
+ADD deployments/bootstrap/* /app/bootstrap/
+COPY --from=build \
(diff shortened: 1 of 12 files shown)


Dockerfile
```diff
@@ -24,5 +24,14 @@ WORKDIR /go/src/github.com/insolar/insolar
 RUN  set -eux; \
      groupadd -r insolar --gid=999; \
      useradd -r -g insolar --uid=999 --shell=/bin/bash insolar
-COPY --from=build /go/src/github.com/insolar/insolar/bin/insolar /go/src/github.com/insolar/insolar/bin/insolard /go/src/github.com/insolar/insolar/bin/keeperd /go/src/github.com/insolar/insolar/bin/pulsard /usr/local/bin/
+COPY --from=build /go/src/github.com/insolar/insolar/application/api/spec/api-exported.yaml /app/api-exported.yaml
```
yellowmegaman: I would prefer to have it mounted as a volume from ConfigMap and not to be placed here at build stage. What do you think?

Dockerfile
```diff
@@ -24,5 +24,14 @@ WORKDIR /go/src/github.com/insolar/insolar
...
 
+# add script and configs required for network bootstrap
+ADD scripts/kube/bootstrap/* /app/bootstrap/
```
yellowmegaman: same

Dockerfile
nordicdyno: it's just binary asset which required and no reason to not put it outside image, probably best option to embed it inside binary (jira task?)

Dockerfile
nordicdyno: Agreed, but we can left it here if it's not critical and create task to fix it, or if it's too nasty I can move this to `kube` repo. Your thoughts?

yellowmegaman: sounds reasonable

Dockerfile
yellowmegaman: yeah, sure

PR #3253, opened 2020-03-12
PENV-44: API for schema

Vany:
**- What I did**
`SchemaService.get` now returns content of api schema 

**- How I did it**
easy

**- How to verify it**
request `SchemaService.get`

Installation now reports version and api of contracts installed

### DIFF
--- a/api/admincontract.go
+++ b/api/admincontract.go
@@ -29,7 +29,6 @@ func NewAdminContractService(runner *Runner) *AdminContractService {
 	}
 	allowedMethods["contract.registerNode"] = true
 	allowedMethods["contract.getNodeRef"] = true
-	allowedMethods["schema.get"] = true
 
 	return &AdminContractService{runner: runner, allowedMethods: allowedMethods}
 }
--- a/api/admincontract.go
+++ b/api/admincontract.go
@@ -29,6 +29,8 @@ func NewAdminContractService(runner *Runner) *AdminContractService {
 	}
 	allowedMethods["contract.registerNode"] = true
 	allowedMethods["contract.getNodeRef"] = true
+	allowedMethods["schema.get"] = true
+
 	return &AdminContractService{runner: runner, allowedMethods: allowedMethods}
 }
 
--- a/api/contract_test.go
+++ b/api/contract_test.go
@@ -89,6 +89,7 @@ func TestTimeoutSuite(t *testing.T) {
 		Options{
 			ContractMethods: map[string]bool{
 				"contract.registerNode": true,
+				"schema.get":            true,
 			},
 		},
 	)
--- a/api/contract_test.go
+++ b/api/contract_test.go
@@ -89,7 +89,6 @@ func TestTimeoutSuite(t *testing.T) {
 		Options{
 			ContractMethods: map[string]bool{
 				"contract.registerNode": true,
-				"schema.get":            true,
 			},
 		},
 	)
--- a/api/main.go
+++ b/api/main.go
@@ -128,7 +128,7 @@ func (ar *Runner) registerAdminServices(rpcServer *rpc.Server) error {
 
 	err = rpcServer.RegisterService(NewSchemaService(ar), "schema")
 	if err != nil {
-		return errors.Wrap(err, "[ registerServices ] Can't RegisterService: funcTestContract")
+		return errors.Wrap(err, "[ registerServices ] Can't RegisterService: schema")
 	}
 
 	return nil
--- a/api/main.go
+++ b/api/main.go
@@ -126,9 +126,9 @@ func (ar *Runner) registerAdminServices(rpcServer *rpc.Server) error {
 		return errors.Wrap(err, "[ registerServices ] Can't RegisterService: funcTestContract")
 	}
 
-	err = rpcServer.RegisterService(NewSchemaService(ar), "schema")
+	err = rpcServer.RegisterService(NewSpecService(ar), "spec")
 	if err != nil {
-		return errors.Wrap(err, "[ registerServices ] Can't RegisterService: schema")
+		return errors.Wrap(err, "[ registerServices ] Can't RegisterService: spec")
 	}
 
 	return nil
--- a/api/main.go
+++ b/api/main.go
@@ -126,7 +126,7 @@ func (ar *Runner) registerAdminServices(rpcServer *rpc.Server) error {
 		return errors.Wrap(err, "[ registerServices ] Can't RegisterService: funcTestContract")
 	}
 
-	err = rpcServer.RegisterService(NewSchemaService(ar), "funcTestContract")
+	err = rpcServer.RegisterService(NewSchemaService(ar), "schema")
 	if err != nil {
 		return errors.Wrap(err, "[ registerServices ] Can't RegisterService: funcTestContract")
 	}
--- a/api/main.go
+++ b/api/main.go
@@ -125,6 +125,12 @@ func (ar *Runner) registerAdminServices(rpcServer *rpc.Server) error {
 	if err != nil {
 		return errors.Wrap(err, "[ registerServices ] Can't RegisterService: funcTestContract")
 	}
+
+	err = rpcServer.RegisterService(NewSchemaService(ar), "funcTestContract")
+	if err != nil {
+		return errors.Wrap(err, "[ registerServices ] Can't RegisterService: funcTestContract")
+	}
+
 	return nil
 }
 
--- a/api/schema.go
+++ b/api/schema.go
@@ -1,0 +1,51 @@
+package api
+
+import (
+	"fmt"
+	"net/http"
+	"os"
+
+	"github.com/insolar/insolar/api/instrumenter"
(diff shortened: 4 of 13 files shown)


api/schema.go
```diff
@@ -0,0 +1,51 @@
+package api
```
LinnikD: license

api/schema_test.go
```diff
@@ -0,0 +1,65 @@
...
+			return insolar.Pulse{}, pulse.ErrNotFound
+		case 2:
+			return insolar.Pulse{}, errors.New("fake error")
+		default:
+			return *insolar.GenesisPulse, nil
+		}
+	})
+	cfg := configuration.NewAPIRunner(false)
+	cfg.Address = "localhost:19192"
+	cfg.SwaggerPath = "testdata/api-exported.yaml"
```
LinnikD: pls add case with no .yaml provided

api/main.go
```diff
@@ -125,6 +125,12 @@ func (ar *Runner) registerAdminServices(rpcServer *rpc.Server) error {
 	if err != nil {
 		return errors.Wrap(err, "[ registerServices ] Can't RegisterService: funcTestContract")
 	}
+
+	err = rpcServer.RegisterService(NewSchemaService(ar), "funcTestContract")
```
LinnikD: funcTestContract?

api/main.go
```diff
@@ -125,6 +125,12 @@ func (ar *Runner) registerAdminServices(rpcServer *rpc.Server) error {
...
+	if err != nil {
+		return errors.Wrap(err, "[ registerServices ] Can't RegisterService: funcTestContract")
```
LinnikD: funcTestContract?

applicationbase/builtin/initialization.go
```diff
@@ -11,6 +11,7 @@ package builtin
 import (
 	"github.com/pkg/errors"
 
+	domain "github.com/insolar/insolar/applicationbase/builtin/contract/domain"
```
LinnikD: domain is not needed in this pr

api/schema_test.go
Vany: it will die before in validator initialization

api/schema_test.go
```diff
@@ -0,0 +1,65 @@
+package api
```
LinnikD: also add functest pls

Vany: It is a serious question, do we need to functest this. 
I belive this mechanism will be obsoleted in next version of platform.
So, if we do not want to develop it, may be we have no need to invest in testing?

api/schema_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	body := rpc.RequestBody{Raw: []byte(`{}`)}
+
+	t.Run("success", func(t *testing.T) {
+		availableFlag = true
+		reply := MAP{}
+
+		err := s.Get(&http.Request{}, &SeedArgs{}, &body, &reply)
+		require.Nil(t, err)
+		require.IsType(t, "string", reply["openapi"], "right openapi")
```
megge-dream: Final argument in this function will be appended to the message the assertion method outputs, if check fails. If you want to use it, return smth like `reply["openapi" must be string`.

api/schema_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+func TestNewSchemaService(t *testing.T) {
+	cfg := configuration.NewAPIRunner(false)
+	cfg.SwaggerPath = "mamamylaramu"
+	runner := Runner{
+		cfg: &cfg,
+	}
+
+	defer func() {
+		require.NotNil(t, recover(), "panic on unexistent file")
```
megge-dream: Final argument in this function will be appended to the message the assertion method outputs, if check fails. If you want to use it, return smth like `expect this func to panic, because file doesn't exist`.

api/main.go
```diff
@@ -125,6 +125,12 @@ func (ar *Runner) registerAdminServices(rpcServer *rpc.Server) error {
 	if err != nil {
 		return errors.Wrap(err, "[ registerServices ] Can't RegisterService: funcTestContract")
 	}
+
+	err = rpcServer.RegisterService(NewSchemaService(ar), "schema")
+	if err != nil {
+		return errors.Wrap(err, "[ registerServices ] Can't RegisterService: funcTestContract")
```
megge-dream: `funcTestContract ` -> `schema `

api/schema.go
```diff
@@ -37,10 +35,50 @@ func NewSchemaService(ar *Runner) *SchemaService {
...
+func Yaml2Json(in interface{}) interface{} {
+	switch i := in.(type) {
+	case map[interface{}]interface{}:
+		ret := map[string]interface{}{}
+		for k, v := range i {
+			switch ii := k.(type) {
+			case string:
+				ret[ii] = Yaml2Json(v)
+			default:
+				log.Panicf("TYPE1 %#v", k)
```
megge-dream: It would be nice to have more info in log message, like what type we expected to get.

api/schema.go
```diff
@@ -37,10 +35,50 @@ func NewSchemaService(ar *Runner) *SchemaService {
...
+			for k, v := range i {
+				ret[k] = Yaml2Json(v)
+			}
+			return ret
+		}
+
+	case int, uint, byte, []byte, string, bool:
+		return i
+	default:
+		log.Panicf("TYPE2 %#v", in)
```
megge-dream: It would be nice to have more info in log message, like what types we expected to get.

application/api/spec/api-exported.yaml
```diff
@@ -846,6 +846,18 @@ paths:
                 oneOf:
                   - $ref: '#/components/schemas/definitions-responses-Status-yaml'
                   - $ref: '#/components/schemas/response-unsignedRequestError'
+  '/admin-api/rpc#schema.get':
```
megge-dream: Did you make PR to spec repo for this change?

application/functest/schema_test.go
```diff
@@ -0,0 +1,46 @@
...
+	err := json.Unmarshal(body, &res)
+	return res, err
+}
+
+func TestSchemaServiceGet(t *testing.T) {
+	requester.SetVerbose(true)
+	ret, err := MakeRPCRequest(t, "schema.get", map[string]interface{}{})
+	require.NoError(t, err)
+
+	log.Panicf(">>> %#v", ret)
```
megge-dream: Why panic at test?

cmd/insolard/api.go
```diff
@@ -64,6 +64,7 @@ func initAPIOptions() (api.Options, error) {
 		"migration.getAddressCount":  true,
 		"deposit.migration":          true,
 		"member.getBalance":          true,
+		"service.get":                true,
```
megge-dream: You don't need this

api/admincontract.go
```diff
@@ -29,6 +29,8 @@ func NewAdminContractService(runner *Runner) *AdminContractService {
 	}
 	allowedMethods["contract.registerNode"] = true
 	allowedMethods["contract.getNodeRef"] = true
+	allowedMethods["schema.get"] = true
```
megge-dream: `schema.get` is not a contract method, so you don't need to add it here

api/contract_test.go
```diff
@@ -89,6 +89,7 @@ func TestTimeoutSuite(t *testing.T) {
 		Options{
 			ContractMethods: map[string]bool{
 				"contract.registerNode": true,
+				"schema.get":            true,
```
megge-dream: `schema.get` is not a contract method, so you don't need to add it here

api/schema_test.go
Tsovak: how do you prove that it works without tests? 
we can remove or not. we'll make a decision later. I really think that we have to develop a test.

api/main.go
Vany: Please use code suggestion next time

api/schema.go
```diff
@@ -37,10 +35,50 @@ func NewSchemaService(ar *Runner) *SchemaService {
...
+
+	case map[string]interface{}:
+		ret := i
+		for k, v := range i {
+			ret[k] = Yaml2Json(v)
+		}
+		return ret
+
+	case []interface{}:
+		{
```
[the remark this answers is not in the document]
Vany: no. It is different types inside of looop

api/schema.go
Vany: ok

application/functest/schema_test.go
Vany: idk

api/admincontract.go
Vany: ok

application/api/spec/api-exported.yaml
Vany: nope

api/schema_test.go
SergeyRadist: Or you can use require.Panics

Vany: > Or you can use require.Panics

interesting, but i want to have additional information

PR #3260, opened 2020-03-20
Penv 122 remove goplugins

megge-dream:
**- What I did**

- Remove goplugins, insgorund and api method for upload contracts.
- Comment func tests - they will be rewritten later.
- Remove `builtin` param from LR config - we always use builtin contracts, no need to set specific param for it.

**- How I did it**

**- How to verify it**


Dockerfile
```diff
@@ -16,8 +16,7 @@ RUN BUILD_NUMBER=${BUILD_NUMBER} \
     BUILD_DATE=${BUILD_DATE} \
     BUILD_TIME=${BUILD_TIME} \
     BUILD_HASH=${BUILD_HASH} \
-    BUILD_VERSION=${BUILD_VERSION} \
-    make insgorund build
+    BUILD_VERSION=${BUILD_VERSION}
```
Tsovak: actually, after deletion make command we don't need this argument. 
I suppose to you deleted extra `make build `stage

applicationbase/functest/pressure_test.go
```diff
@@ -7,239 +7,239 @@
...
-			require.Equal(t, float64(1), bal)
-			time.Sleep(1 * time.Second)
-		}
-	})
-}
+// import (
+// 	"sync"
+// 	"sync/atomic"
+// 	"testing"
+// 	"time"
```
Tsovak: did we comment it to rewrite/fix latter?

applicationbase/functest/test_utils.go
```diff
@@ -81,143 +77,3 @@ func createMember(t *testing.T) *CommonUser {
 	member.Ref = ref
 	return member
 }
-
```
Tsovak: after deletion I thing `var contracts = map[string]*contractInfo{}` no more need and can be deleted

applicationbase/functest/pressure_test.go
megge-dream: yes, we will rewrite it!

applicationbase/functest/test_utils.go
megge-dream: you are absolutely right!

Dockerfile
megge-dream: changed it

cmd/insgocc/insgocc.go
```diff
@@ -77,8 +77,6 @@ func (r *machineTypeFlag) Set(arg string) error {
 		fallthrough
 	case "go":
 		fallthrough
-	case "golang":
```
Vany: "go" will fall through too

api/main.go
```diff
@@ -12,7 +12,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/insolar/x-crypto"
+	crypto "github.com/insolar/x-crypto"
```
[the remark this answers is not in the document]
megge-dream: without alias `go import` changes it to `crypto` lib

PR #3263, opened 2020-03-26
Ops 111 bootstrap

yellowmegaman:
**- What I did**
Added option to generate sane names for certificates
**- How I did it**
Added `--propername` flag for bootstrap subcommand
**- How to verify it**
run insolar bootstrap -p true

By default propername = false, so it won't affect existing setups.


yellowmegaman:
image

applicationbase/bootstrap/keys.go
```diff
@@ -89,8 +90,14 @@ func createKeysInDir(
 			return nil, errors.Wrap(err, "[ createKeysInDir ] couldn't marshal keys")
 		}
 
-		inslogger.FromContext(ctx).Info("Genesis write key " + filepath.Join(dir, keyname))
-		err = makeFileWithDir(dir, keyname, result)
+		if properName {
```
SergeyRadist: You can di like this
if properName {
    keyname = n.CertName
}
and then log and makeFileWithDir once

yellowmegaman: Nice! Thanks

yellowmegaman: Done

PR #3267, opened 2020-04-02
MN-127: move long and nightly CI test scripts to makefile

pavel-moskovkin:
Scripts used in long and nightly TeamCity configurations partially added as a part of Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,10 +26,13 @@ FUNCTEST_COUNT ?= 1
 TESTED_PACKAGES ?= $(shell go list ${ALL_PACKAGES} | grep -v "${MOCKS_PACKAGE}")
 COVERPROFILE ?= coverage.txt
 TEST_ARGS ?= -timeout 1200s
+TEST_ARGS_NIGHTLY ?= -timeout 1000m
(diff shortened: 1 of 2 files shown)


pavel-moskovkin:
Correspondence of existing TC jobs to the new (or already existing) targets in makefile:
1 Long Tests
  1.1 Unit Tests  			ci-test-unit-long
  1.2 Slow Tests   			ci-test-slow-long-repeat
  1.3 Func Tests  			ci-test-func-long
  1.4 Integration Tests  	  	ci-test-integrtest-long-repeat
2 Nightly Tests
  2.1 Func Tests  			ci-test-func-nightly
  2.2 Integration tests  	  	ci-test-integrtest-nightly
  2.3 Slow Tests 			ci-test-slow-nightly-repeat
  2.4 Unit Tests 			ci-test-unit-nightly

twwlf:
Maybe we should drop "ci_" prefixed targets and use same targets for local and CI runs instead? Additional flags like "-json" can be replaced by env variables, which, in their turn, would be defined on CI.

PR #3273, opened 2020-04-15
PENV-3: use insconfig

Vany:
**- What I did**
use insconfig in pulsard
**- How I did it**

**- How to verify it**


configuration/configuration.go
```diff
@@ -80,7 +80,7 @@ func NewGenericConfiguration() GenericConfiguration {
 // NewPulsarConfiguration creates new default configuration for pulsar
 func NewPulsarConfiguration() PulsarConfiguration {
 	return PulsarConfiguration{
-		Log:      NewLog(),
```
twwlf, line 83: don't break formatting please (or at least keep it consistent)

configuration/configuration.go
```diff
@@ -80,7 +80,7 @@ func NewGenericConfiguration() GenericConfiguration {
...
+		Log: NewLog(),
 		Pulsar:   NewPulsar(),
 		Tracer:   NewTracer(),
 		KeysPath: "./",
```
twwlf, line 86: Is it a default value? Would it be used while "KeysPath" isn't present in config file?

cmd/pulsard/main.go
```diff
@@ -54,31 +57,28 @@ func parseInputParams() inputParams {
 
 // Need to fix problem with start pulsar
 func main() {
+	_cfg := configuration.NewPulsarConfiguration()
```
SergeyRadist, line 60: Why do we need NewPulsarConfiguration() ?
Can we use struct directly, like cfg := PulsarConfiguration{} ?

Vany: because we are using accessors instead of direct access

configuration/configuration.go
Vany: nope

configuration/configuration.go
Vany: ok

cmd/pulsard/main.go
SergeyRadist: my point was that NewPulsarConfiguration was added as constructor with defaults, now we don't need it. But for beauty, it's ok )

megge-dream: I think using `NewPulsarConfiguration` instead of ` PulsarConfiguration{}` is confusing - as in comment above says it's not clear are we using default value in config or not. May be we should use direct access for clarity?

PR #3288, opened 2020-06-23
MN-467: add an optional JWT validation

MikhailPopov:
**- What I did**
- I have added authentication properties to the exporter config.
- Now I configure a GRPC server with an interceptor if required.
- The interceptor finds Authorization metadata and validates a JWT contained in it.

**- How to verify it**
run make test_all


configuration/export.go
```diff
@@ -11,11 +11,28 @@ package configuration
 type Exporter struct {
```
Vany, line 11: file have name export, why sthis object have different name?

configuration/export.go
```diff
@@ -11,11 +11,28 @@ package configuration
...
 	// Addr specifies address where exporter server starts
 	Addr string
+	Auth Auth
+}
+
+// Auth specifies parameters for a token-based authorization of an observer
+type Auth struct {
```
Vany, line 18: annotate it like this pls: https://github.com/insolar/block-explorer/blob/master/configuration/configuration.go

configuration/export.go
MikhailPopov: I haven't any idea, why it was before, but I will rename the file.

configuration/export.go
MikhailPopov: @Vany The Insolar platform code does not use the "YamlTemplater" from the insconfig library. It will be useless work to cover an exporter config with an insconfig annotations.
I think this work should be done in a separate task for the whole configuration and adding the YamlTemplater.

configuration/export.go
MikhailPopov: File has been renamed.

Vany: thx

PR #3314, opened 2020-07-30
MN-631: validate contract version

sansaian:
**- What I did**
Validate contract version on heavy side

**- How I did it**
add version in helper contract
**- How to verify it**
starting unit test or start insolar and observer


server/internal/heavy/components.go
```diff
@@ -872,7 +874,12 @@ func newComponents(
 	genesisOptions genesis.Options,
 	genesisOnly bool,
 	apiOptions api.Options,
+	contractVersion int64,
 ) (*components, error) {
+	if contractVersion <= 0 {
+		log.Fatalf("incorrect allowed contract version application: %v", contractVersion)
```
megge-dream, line 880: use `logger := inslogger.FromContext(ctx)` instead of `log`

PR #3329, opened 2020-09-24
NOISSUE: remove vendor, update go to 1.15

MikhailPopov:
**- What I did**
I've tried to remove the vendor directory from the repository and update versions of go to 1.15 and ugorji/go/codec to 1.1.8


MikhailPopov:
The experiment failed because of the ugorji/go/codec library when we tried to use it in the Mainnet and Observer. In the next PR, we didn't change the library's version https://github.com/insolar/insolar/pull/3330.

